PR #44, opened 2026-05-04
fix(core): add envDir to load .env from user project root

hanpham32:
envDir isn't set so Vite couldn't scan user project's environment variables. This change allows users to use .env / .env.local with VITE_ prefixed variables in their project.


1weiho:
LGTM, thanks

1weiho (approved):

PR #53, opened 2026-05-04
Fix inspect overlay measurement during panel open

chentyke:
## Summary
- Rework the inspect overlay to keep a live selected DOM anchor instead of caching a click-time `DOMRect`.
- Remeasure the highlight frame after layout changes while the inspector panel opens, and on root/target resize.
- Add a patch changeset for `@open-slide/core`.

## Root cause
The first element selection opens the inspector panel. While that panel expands from zero width, the slide viewport changes size and `SlideCanvas` recomputes its fit scale and centered position. The previous overlay measured `getBoundingClientRect()` before that layout settled, so the highlight frame could render from stale geometry. Once the panel was already open, the layout was stable, which is why later selections lined up correctly.

## Validation
- `pnpm exec biome check packages/core/src/app/components/inspector/inspect-overlay.tsx .changeset/silver-inspect-frames.md`
- `pnpm test packages/core/src/app/lib/utils.test.ts packages/core/src/app/lib/sdk.test.ts`
- `pnpm core build`
- Browser-verified first-open inspect selection on the Google history demo deck.

## Notes
- `pnpm check` currently fails on unrelated existing repository issues in `apps/web`, `apps/demo`, and `packages/cli`.
- `pnpm core typecheck` currently fails on an unrelated existing readonly ref assignment in `packages/core/src/app/components/player.tsx`.

### DIFF
--- a/.changeset/silver-inspect-frames.md
+++ b/.changeset/silver-inspect-frames.md
@@ -1,0 +1,5 @@
+---
+'@open-slide/core': patch
+---
+
+Keep the inspect highlight aligned while the inspector panel opens.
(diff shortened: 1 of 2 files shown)


1weiho:
Hi 👋🏻
LGTM, thanks!

1weiho (approved):

PR #54, opened 2026-05-04
fix(core): make dev server work on Windows

Willaiem:
- Normalize `/@fs/` URL so Windows drive-letter paths get the required separator (was producing `/@fsE:/...` and breaking slide loading).
- Explicitly include `react`, `react-dom`, `react-dom/client`, and `next-themes` in `optimizeDeps.include` so they get pre-bundled into ESM. Without this, the in-`node_modules` `root` prevents `@vitejs/plugin-react`'s auto-include from firing and Vite serves the raw CJS files, which lack the named exports the app imports.

### DIFF
--- a/.changeset/fix-windows-dev-server.md
+++ b/.changeset/fix-windows-dev-server.md
@@ -1,0 +1,5 @@
+---
+'@open-slide/core': patch
+---
+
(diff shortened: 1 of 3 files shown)


1weiho:
LGTM, thanks!

PR #113, opened 2026-05-13
fix(core): guard mutation endpoints against cross-origin requests

DeryFerd:
## Summary

This PR addresses two validated security findings from the latest local security scan:

- FD-002: Unprotected `POST /__edit` allows unauthorized local source mutation.
- FD-003: Unprotected `POST /__comments/add` allows unauthorized marker injection.

Both endpoints can write directly to slide source files through `fs.writeFile`, but previously accepted mutating requests without checking caller trust. This change introduces request validation at the middleware boundary and adds focused regression tests.

## Context and Risk

The findings were validated in the scan artifacts generated on May 13, 2026 (`.security-scans/open-slide/5a39c31_20260513_202034`). The issue is straightforward:

- `POST /__edit` can rewrite slide source files based on request body operations.
- `POST /__comments/add` can insert persistent markers into slide files.
- `DELETE /__comments/:id` can remove existing markers.

Before this fix, these mutating routes did not enforce origin trust or request metadata checks. In practice, that means if a browser or network client can reach the local dev server, it can attempt write operations against project files.

In a normal dev setup this is often local-only exposure, but it is still a real boundary. Browser-based local-target requests, accidental port exposure, reverse proxy misconfiguration, or shared dev environments can turn this into a practical risk.

### 1) Added a shared mutation request guard

`packages/core/src/vite/comments-plugin.ts`

A new helper `validateMutationRequest(req, opts)` now validates mutating requests before parsing bodies or touching files.

The guard enforces:

- JSON content type for routes that expect JSON payloads (`application/json` prefix accepted, including charset suffixes).
- Rejects `Sec-Fetch-Site: cross-site` with `403`.
- Rejects `Origin: null` (opaque origin) with `403`.
- Rejects malformed `Origin` headers with `403`.
- Validates that `Origin` matches effective server origin built from protocol + host.
- Supports forwarded deployments by honoring `x-forwarded-proto` and `x-forwarded-host` when present.
- Returns structured failures with explicit status and error message.

### 2) Applied guard to all comment/edit write paths in this plugin

Guarded routes:

- `POST /__edit` (covers `/` and `/batch` under this middleware mount)
- `POST /__comments/add`
- `DELETE /__comments/:id`

For `POST` write routes, JSON content type is required.
For `DELETE`, origin/fetch metadata checks apply without content-type requirement.

### 3) Added regression tests for trust validation logic

`packages/core/src/vite/comments-plugin.test.ts`

New test block: `describe('validateMutationRequest', ...)`

Coverage includes:

- Accept same-origin JSON mutation requests.
- Accept forwarded host/proto when origin matches proxy headers.
- Reject non-JSON mutation bodies when JSON is required.
- Reject cross-site fetch metadata.
- Reject opaque origins.
- Reject origin/host mismatch.
- Allow trusted requests when `Origin` is absent (for compatibility with non-browser/local tooling that may omit it).

## Why This Fix Is Structured

- The trust check sits at the middleware entry point, not deeper in business logic. That keeps the enforcement close to the attack surface and avoids route-specific drift.
- A shared helper makes behavior consistent across mutating endpoints.
- `x-forwarded-*` support prevents false positives when users run dev behind a local proxy/tunnel.
- The compatibility rule for missing `Origin` is deliberate. Some local tooling and scripted clients omit it. Blocking those outright would break legitimate workflows.
- `Sec-Fetch-Site` is leveraged as an extra browser signal. It is not the only control, but it is useful and low cost.

## Behavior Changes

Expected behavior after this PR:

- Same-origin editor operations continue to work as before.
- Cross-origin browser requests to these mutation endpoints are blocked.
- Requests with invalid or mismatched origin information are blocked.
- JSON mutation endpoints reject non-JSON payloads with `415`.
- Comment deletion now uses the same trust boundary as comment creation and source editing.

## Validation Performed

Command run:

- `corepack pnpm test -- packages/core/src/vite/comments-plugin.test.ts`

Result:

- Passed.
- Vitest summary from run:
  - Test files: 9 passed
  - Tests: 195 passed

This includes the new mutation-request guard regression tests and existing related plugin tests.

## Scope Notes

This PR intentionally targets FD-002 and FD-003 only, as requested. It does not refactor unrelated middleware. The change is narrow and localized to `comments-plugin.ts` plus matching test coverage.

## Reviewer Guide

If you want a fast review path:

1. Read `validateMutationRequest` in `comments-plugin.ts`.
2. Confirm guard invocation at:
   - `POST /__edit`
   - `POST /__comments/add`
   - `DELETE /__comments/:id`
3. Check the new `validateMutationRequest` test block in `comments-plugin.test.ts`.
4. Verify no functional edit/comment flows were rewritten beyond trust gating.

## Follow-up Ideas (Optional, not in this PR)

- Add per-session CSRF/auth token for mutation routes for stronger non-browser client enforcement.
- Reuse the same guard pattern across other write-capable dev middleware surfaces (`__notes`, `__design`, etc.) for consistency.

## Summary by CodeRabbit

* **Bug Fixes**
  * Mutation endpoints now include request validation for edit and comment operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/1weiho/open-slide/pull/113?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/patch-core-mutation-guard.md
+++ b/.changeset/patch-core-mutation-guard.md
@@ -1,0 +1,5 @@
+---
+'@open-slide/core': patch
+---
+
+Harden comments and edit mutation endpoints by validating request origin/fetch metadata and requiring JSON content-type for write routes.
--- a/packages/core/src/vite/comments-plugin.test.ts
+++ b/packages/core/src/vite/comments-plugin.test.ts
@@ -5,8 +5,8 @@ import {
   b64urlEncode,
   parseMarkers,
   safeAssetIdentifier,
-  validateMutationRequest,
 } from './comments-plugin.ts';
+import { validateMutationRequest } from './request-guard.ts';
 
 describe('b64url encoding', () => {
   it('round-trips arbitrary unicode strings', () => {
--- a/packages/core/src/vite/comments-plugin.test.ts
+++ b/packages/core/src/vite/comments-plugin.test.ts
@@ -5,6 +5,7 @@ import {
   b64urlEncode,
   parseMarkers,
   safeAssetIdentifier,
+  validateMutationRequest,
 } from './comments-plugin.ts';
 
 describe('b64url encoding', () => {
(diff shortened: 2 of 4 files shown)


1weiho:
Hi @DeryFerd 👋🏻
Thanks for the fix, guard logic and tests look great.

One suggestion before merge: could we move `validateMutationRequest` (and its helpers) out to something like `packages/core/src/vite/request-guard.ts` in this PR?

Your follow-up section already notes the same issue exists on /__notes, /__design, /__slides, /__assets, /__folders. Extracting now keeps those follow-ups to a 1-line import, and avoids unrelated plugins importing from a comments module later.

DeryFerd:
Hello @1weiho !

Thanks for the suggestion — great call.

I extracted `validateMutationRequest` and its helper logic into a dedicated module at `packages/core/src/vite/request-guard.ts`, then updated `comments-plugin.ts` to import from it.

I also updated the existing guard tests to import `validateMutationRequest` from the new module, so coverage stays the same and behavior is unchanged.

Validation run:
- `corepack pnpm test -- packages/core/src/vite/comments-plugin.test.ts`
- Result: 9 files passed, 195 tests passed.

1weiho:
@DeryFerd LGTM, thanks!

1weiho (approved):

PR #154, opened 2026-05-23
fix(core): truncate long slide titles in the editor header

ridemountainpig:
## Summary
- Long deck titles overflowed the editor header and overlapped the right-side controls (Design / Inspect / Present).
- Swapped the absolute-positioned title overlay for a 3-zone flex layout: side zones are `shrink-0`, the title sits in a `flex-1 min-w-0` middle zone so `truncate` actually engages.
- Dropped the brittle `max-w-[min(34rem,calc(100vw-22rem))]` formula and the `pointer-events-none/auto` workaround that the absolute overlay needed.

## Before / after
At ~900px viewport with a long title, the title used to render in full on top of the Design / Inspect buttons. It now truncates with an ellipsis and leaves a clean gap to the right-side controls.

https://github.com/user-attachments/assets/12971fb0-85ce-4c34-be8c-2e68e2736a7c

## Test plan
- [x] `pnpm core typecheck`
- [x] `pnpm check`
- [x] Verified at 600 / 900 / 1100 / 1280 / 1900px with a long title — truncation triggers when needed, no overlap at any width.
- [x] Verified rename (pencil → input) still works without overlap.

## Trade-off
The title is now centered between the side zones rather than the viewport. When the right zone is wider than the left (typical), the title shifts a few dozen pixels left of the strict viewport center. This is the necessary trade-off for guaranteeing no overlap.

### DIFF
--- a/.changeset/header-title-truncate.md
+++ b/.changeset/header-title-truncate.md
@@ -1,0 +1,5 @@
+---
+'@open-slide/core': patch
+---
+
+Truncate long slide titles in the editor header instead of letting them overlap the right-side controls.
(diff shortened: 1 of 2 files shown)


1weiho:
@ridemountainpig Hey Yen 👋🏻
LGTM, thanks!

1weiho (approved):

PR #156, opened 2026-05-23
feat(core): editable PPTX export via built-in renderer

Luotee:
## Summary

Activates the reserved **"Export as PPTX"** dropdown item (previously a *Coming soon* placeholder) with a real, **editable** PowerPoint export — native text and shapes, not a rasterised image. It sits alongside the existing *Export as image PPTX*.

Per our discussion, the rendering logic is **built directly into `@open-slide/core`** rather than shipped as a third-party package. This addresses both concerns raised on the original revision of this PR:

- **Security** — no third-party export package to trust; the code lives in this repo and is reviewable here.
- **Maintainability** — the renderer evolves in lockstep with the slide primitives (the deck-bundling stub is updated in the same repo, so primitives like `Step`/`Steps` stay supported).

The renderer measures the laid-out DOM with **Playwright (Chromium)** to stay pixel-faithful to what the browser draws. Playwright is the *only* thing a user installs, and it is an **optional peer dependency**: when it (or its Chromium binary) is missing, the export returns a copy-pasteable install command instead of failing.

> This supersedes the original external-package approach — the engine is now vendored in-repo, so installing `@helping-ai-workflow/slide-to-pptx` is no longer required.

## How it works

- New dev-server route `POST /__os/api/export/pptx` (registered in `api-plugin.ts`, `apply: 'serve'`).
- A preflight probes Playwright. **Missing** → JSON `{status:'playwright-missing'}` with `<pm> add -D playwright && npx playwright install chromium`. **Present** → SSE progress stream, then a base64 `.pptx` `done` event.
- Engine under `packages/core/src/vite/pptx/` (server-only): load deck → render HTML → measure with Chromium → IR → pptxgenjs → OOXML post-process.

## File layout

| Area | Files |
|---|---|
| Render engine (server-only) | `packages/core/src/vite/pptx/*` |
| Dev-server route + handler | `vite/routes/export-pptx{,-handler}.ts`, `pm-detect.ts` |
| Client trigger + toasts | `app/lib/export-pptx-native.ts`, `app/components/pptx-*-toast.tsx` |
| UI wiring | `app/routes/slide.tsx` (activates the reserved slot) |
| Deps / build | `package.json`, `tsdown.config.ts`, scoped `biome.json` override |

## Minimal-impact notes

- `slide.tsx`: only the reserved *Coming soon* slot becomes a working item, mirroring the existing `exportImagePptx`. HTML / PDF / image-PPTX paths are untouched.
- New runtime deps: `esbuild`, `pptxgenjs`, `jszip`, `fast-xml-parser` (server-side only — they never reach the browser bundle). `playwright` is a `devDependency` + an **optional** `peerDependency`, so existing consumers are never forced to download Chromium.
- `biome.json`: one override turns off `noExplicitAny` / `noNonNullAssertion` for `packages/core/src/vite/pptx/**` only (the vendored engine does heavy DOM/OOXML work); no author-owned path is affected.

## Test plan

- `pnpm typecheck` ✅ · `pnpm format:check` + `pnpm lint` ✅ · `pnpm test` ✅ 289/289 (new: `export-pptx-handler.test.ts` incl. path-traversal + playwright-missing, `pm-detect.test.ts`).
- End-to-end on a real consumer (21 decks): every deck exports; **0 shapes overflow** the 1920×1080 canvas; cover titles render exactly once.
- SVG stress fixtures verified in PowerPoint: multi-subpath `<path>` (X / +) no longer draws a connecting segment; filled paths (diamond / circle) render; cubic/quadratic/arc curves intact.
- Playwright-missing path verified end-to-end (returns the correct install prompt).

## Open question

The engine adds a few MB to `@open-slide/core` (pptxgenjs/jszip/etc.); Chromium (~130 MB) stays opt-in via the peer. If you'd prefer to keep core leaner, I'm happy to split the engine into a separate workspace package instead — your call.

## Release Notes

* **New Features**
  * Added an **“Export as PPTX”** option for slide downloads, generating native editable PowerPoint output.
  * Shows live progress while the PPTX export is being created.

* **Bug Fixes**
  * When PPTX export can’t run due to a missing required component, the flow now provides clear, actionable install guidance instead of failing silently.
  * Improved handling of export failures with user-friendly error messaging.

* **Localization**
  * Updated PPTX-related messages for English, Japanese, Simplified Chinese, and Traditional Chinese.
  * Added a **Copy command** label used for the install guidance UI.


1weiho:
Hey @Luotee, thanks for this PR! The implementation is really well done, and I appreciate the effort you put into it.

I do want to raise one architectural concern before we move forward. I'd prefer to have the PPTX rendering logic built directly into open-slide rather than depending on an external package (`@helping-ai-workflow/slide-to-pptx`), for two reasons:

1. **Security**: As a framework used by people, asking users to install a third-party package for a core export feature adds a dependency we can't fully audit or control.
2. **Maintainability**: When open-slide makes breaking changes to slide primitives, we need to update all export paths together. An external plugin creates a sync gap we can't fix on our end.

Would you be open to co-working on this as a built-in feature? I'd love to collaborate on landing it together. Let me know what you think!

Luotee:
Hi @1weiho , thanks for the kind words and the thoughtful concerns. Both points totally make sense.

One thing worth flagging though: even if we merge this directly into open-slide, users will still need to run `npx playwright install chromium` to use the export. The PPTX rendering precision currently relies on a real browser, so the extra download is hard to avoid with the current approach.

That said, I think it's worth exploring whether we can hit the same precision without the browser dependency. Would make the whole thing cleaner. I'm traveling this week, so let me poke at this when I'm back and we can sync on a plan then. Excited to collaborate on landing this properly! 🙌

Luotee:
Hi Yiwei — following the direction we settled on, I've reworked this PR. The main changes:

**Built-in engine, no external package**
It previously rendered through the third-party `@helping-ai-workflow/slide-to-pptx` package; the whole export engine is now **built directly into `@open-slide/core`**. That addresses both points you raised:
- **Security** — there's no third-party export package to trust; the code lives in this repo and is reviewable here.
- **Maintainability** — the engine is maintained in the same repo as the slide primitives, so they can't drift apart (it even supports newer primitives like `Step`/`Steps`).

**Playwright as the only optional dependency**
Measuring the laid-out DOM still needs Playwright (Chromium) so the exported pptx stays pixel-faithful to the browser. I've made it an **optional peer dependency**: existing users are never forced to download Chromium; if it's missing, clicking "Export as PPTX" surfaces a copy-pasteable install command instead of erroring; once installed, it exports an **editable, native PPTX** (real text/shapes, not an image). It lives alongside your existing image-PPTX export — the two are complementary.

I've wired it up to the "Coming soon" PPTX menu item you'd reserved.

**Verification:** all 21 of my test decks export cleanly with 0 elements overflowing the slide; the SVG stress fixtures (multi-subpath paths, filled paths, curves) all render correctly; `typecheck` / `lint` / `test` (289) are green.

Details are in the updated PR description. Would love your review — happy to adjust anything! 🙌

Luotee:
Thanks @coderabbitai — addressed in 321aef2.

**Fixed**
- **bun.lock detection** — `pm-detect` now matches `bun.lock` (current text default), not only `bun.lockb`.
- **Fragment instrumentation** — `walk()` now descends into `React.Fragment`, so primitives inside `<>…</>` get tagged.
- **Unbounded ESM cache** — the slide bundle is imported via a content-addressed URL, so repeated exports reuse one module record instead of leaking one per export in the dev server.
- **Grouping** — a non-contiguous group key is now skipped instead of aborting grouping for the rest of the slide.
- **Clipboard toast** — guarded against a missing Clipboard API; success is shown only after the write resolves.
- **Transforms / off-canvas leaves** — rotated/scaled elements now fall back to an image; a wholly off-canvas overflow leaf gets a transparent placeholder instead of emitting off-slide.

A few comments are stale — they reference the pre-force-push revision (the external `@helping-ai-workflow/slide-to-pptx` peer dep, the old changeset filename, and a `Date.now()` toast id that no longer exists). The progress-toast `default`-branch note is in an upstream file this PR doesn't touch.

While verifying I also closed two text-fidelity gaps this surfaced: an inline `<svg>`/`` icon was suppressing its sibling label text (e.g. the "▲ Vercel Labs" header), and CSS `text-transform` / `letter-spacing` weren't applied — both render faithfully now.

All green: 21 decks export with **0 canvas overflow**, `typecheck` / `lint` / `test` (290) pass.

Luotee:
Thanks @coderabbitai — addressed the atomic-write race in 562f6a2: the slide bundle is now written to a unique temp file and `rename`d into place (atomic on the same filesystem), so two concurrent exports of the same deck can't leave `import()` reading a half-written file. The earlier review items are resolved.

PR #169, opened 2026-05-26
fix(core): preserve backgrounds in PDF export

### LINKED ISSUE #168: [Bug]: Pdf export loose background image or gradients
### Summary

I've got a slideshow presentation with some slide having some picture as background or css gradient in it and some white text hover it.
When I export the slideshow to pdf, these background image and gradient are loose and are just plain white background. Text being white, white on white I see nothing...

### Steps to reproduce

1. Create a presentation with a background image or gradient image
2. Export it to pdf
3. Observe that the background image or gradient are not pres

### Expected behaviour

background image or gradient should be exported in pdf

### Actual behaviour

background image or gradient should are not exported in pdf

### Affected package

scaffolded project

### Version

@open-slide/core@1.6.0

### Environment

Node v24.13.0, npm 11.6.2, macOS 26.5, Chrome 148.0.7778.179

### Additional context

_No response_

chentyke:
Fixes #168

## What

Preserves authored `background-image` values when exporting slides to PDF.

- Keep CSS gradients in the print render tree.
- Keep layered backgrounds such as `linear-gradient(...), url(...)`.
- Continue stripping `box-shadow` in the print container to avoid expensive PDF soft-mask output.

## Why

The PDF exporter was neutralizing any element whose computed `backgroundImage` contained `gradient(` before calling `window.print()`. That turned the entire inline background into `none`, so layered backgrounds lost both the gradient and image layer. The print frame then fell back to white, making white text disappear on affected slides.

## How

- Remove the inline gradient cleanup pass before `window.print()`.
- Remove the print CSS rule that cleared pseudo-element background images.
- Narrow the existing print optimization comment to `box-shadow`, which is still stripped.

## Screenshots

| Before | After |
| --- | --- |
| Before: PDF background stripped | After: PDF background preserved |

## Test plan

- [x] `pnpm --filter @open-slide/core typecheck`
- [x] `pnpm --filter @open-slide/core build`
- [x] Manual repro with a slide containing a CSS gradient background
- [x] Manual repro with a layered `linear-gradient(...), url(...)` background
- [x] Verified the print DOM keeps both background values before PDF generation

## Reviewer notes

The original optimization was intended to avoid slow PDF rendering for gradients in macOS Preview, but it removed user-authored backgrounds. This keeps the shadow optimization while preserving slide visuals.

## Summary by CodeRabbit

* **Bug Fixes**
  * PDF export now preserves gradient and image backgrounds when exporting slides, ensuring backgrounds appear correctly in generated documents.
  * Improved handling of print styling to retain background images while preventing unwanted shadow artifacts in exported PDFs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/1weiho/open-slide/pull/169?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/preserve-pdf-backgrounds.md
+++ b/.changeset/preserve-pdf-backgrounds.md
@@ -1,0 +1,5 @@
+---
+"@open-slide/core": patch
+---
+
+Preserve gradient and image backgrounds when exporting slides to PDF.
--- a/packages/core/src/app/lib/export-pdf.ts
+++ b/packages/core/src/app/lib/export-pdf.ts
@@ -63,19 +63,13 @@ const PRINT_STYLES = `
     transform: scale(0.5);
     transform-origin: top left;
   }
-  /* Chromium serializes box-shadow and CSS gradients as PDF transparency
-     groups / soft masks. macOS Preview re-composites those on every page
-     turn, causing 0.5–2s per-page lag. Strip them in the print container
-     only — gradients on pseudo-elements via CSS (DOM walk can't reach them),
-     inline-style gradients via neutralizeGradientBackgrounds() below. */
+  /* Chromium serializes box-shadow as PDF transparency groups / soft masks.
+     macOS Preview re-composites those on every page turn, causing lag. Strip
+     shadows in the print container while preserving authored backgrounds. */
   #${PRINT_ROOT_ID} *,
   #${PRINT_ROOT_ID} *::before,
   #${PRINT_ROOT_ID} *::after {
     box-shadow: none !important;
-  }
-  #${PRINT_ROOT_ID} *::before,
-  #${PRINT_ROOT_ID} *::after {
-    background-image: none !important;
   }
 }
 `;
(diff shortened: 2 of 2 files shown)


1weiho:
Hey Tyke! Thanks for the fix!

I’m still figuring out how to make the PDF export better, because when we keep the CSS gradient in the PDF, opening the file in macOS feels really slow to render, like in this demo.

https://github.com/user-attachments/assets/a69db88f-f530-4e31-9079-73bc3158d8e3

chentyke:
Hey, thanks for taking a look and for the extra context.

That makes sense. Preserving the CSS gradients fixes the missing backgrounds, but it also brings back the macOS PDF rendering slowdown that the previous workaround was trying to avoid.

I'll close this PR for now so it doesn't imply this is the final direction. I'll also take a look and see if there is a better way to preserve slide backgrounds without making the exported PDF painful to open.

Thanks again!

PR #191, opened 2026-06-01
feat(core): toggle slide canvas between 16:9 and 4:3

itskylechung:
## What

Adds a per-slide aspect-ratio toggle so a deck can mix 16:9 and 4:3 slides. A `meta.aspect` field (`'16:9'` default, `'4:3'` opt-in) drives the canvas dimensions everywhere a slide is rendered, and a DEV-only toolbar button flips it.

## How

- **`meta.aspect`** — new optional `SlideAspect` field on `SlideMeta`. `getCanvasDims()` in `sdk.ts` maps it to dimensions (16:9 → 1920×1080, 4:3 → 1440×1080; height stays fixed so existing 1080-tall content isn't rescaled, the 4:3 frame just narrows).
- **Rendering** — `SlideCanvas` takes an `aspect` prop; threaded through the editor canvas, thumbnail rails (vertical + horizontal), presenter, present mode + overview grid, and home-page card previews.
- **Exports** — `export-pdf` (`@page size`, frame + supersample dims) and `export-html` (`.os-frame` size, fit script) read the slide's aspect.
- **Writer** — `setSlideAspect` / `updateMetaAspectInSource` edit `meta.aspect` in `index.tsx` via a new `PUT /__slides/:id/aspect` dev endpoint. Resetting to 16:9 removes the field (it's the default) and cleans up the line so the file doesn't accrete blank lines.
- **Toolbar button** — DEV-only segmented `16:9` / `4:3` control next to the Inspect toggle.
- **i18n** — added strings for en / ja / zh-cn / zh-tw.

## Tests

- New `getCanvasDims` unit tests (default / 16:9 / 4:3).
- New `updateMetaAspectInSource` tests covering insert / replace / remove / no-op / fresh-meta, including the no-stray-blank-line invariant.
- Full suite: 263 passing. `typecheck` + `biome check` clean.

A changeset (`minor`) is included.

## Summary by CodeRabbit

* **New Features**
  * Added an aspect ratio toggle to switch slides between 16:9 and 4:3; choice is stored with the slide and honored across editor, thumbnails, presenter, present mode, and HTML/PDF exports
  * Slide aspect can be updated programmatically via a new slide metadata update route

* **Localization**
  * Added toggle UI text in English, Japanese, Simplified Chinese, and Traditional Chinese

### DIFF
--- a/.changeset/aspect-toggle.md
+++ b/.changeset/aspect-toggle.md
@@ -1,0 +1,5 @@
+---
+"@open-slide/core": minor
+---
+
+Add a toolbar button to toggle a slide's canvas between 16:9 and 4:3. Stored as `meta.aspect`; honored in the editor, presenter, present mode, and HTML/PDF exports.
--- a/packages/core/src/app/components/player.tsx
+++ b/packages/core/src/app/components/player.tsx
@@ -3,7 +3,7 @@ import { useClickPageNavigation } from '@/lib/use-click-page-navigation';
 import { useWheelPageNavigation } from '@/lib/use-wheel-page-navigation';
 import { cn } from '@/lib/utils';
 import type { DesignSystem } from '../lib/design';
-import type { Page } from '../lib/sdk';
+import type { Page, SlideAspect } from '../lib/sdk';
 import type { SlideTransition } from '../lib/transition';
 import { usePrefersReducedMotion } from '../lib/use-prefers-reduced-motion';
 import { PresentBlackoutOverlay } from './present/blackout-overlay';
@@ -31,6 +31,7 @@ type Props = {
   pages: Page[];
   design?: DesignSystem;
   transition?: SlideTransition;
+  aspect?: SlideAspect;
   index: number;
   onIndexChange: (index: number) => void;
   onExit: () => void;
@@ -49,6 +50,7 @@ export function Player({
   pages,
   design,
   transition,
+  aspect,
   index,
   onIndexChange,
   onExit,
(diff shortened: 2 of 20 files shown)


NetanelRotem:
Cross-linking related work, since this PR is the closest thing to it in the repo.

I opened #404 asking for a configurable canvas, and #407 implements it — **as a draft, WIP, explicitly not a merge request.** I would much rather get the API shape right first, and this PR is the obvious place to check that against.

The two overlap but aim at different problems, and I do not think they conflict:

- **This PR is per-slide.** `meta.aspect` on the slide plus a toolbar toggle, so one deck can mix ratios. It holds `height` at 1080 so existing content is not rescaled, which means 16:9 ↔ 4:3 and portrait is out of reach by construction.
- **#407 is workspace-level.** `canvas` in `open-slide.config.ts`, taking a preset or explicit `{ width, height }`, on the premise that a deck has one output format and that is a project setting. It covers 4:5 and 9:16, which is what I actually needed — LinkedIn document carousels.

If both land, the config value is the natural default that a per-slide `meta.aspect` would override, and this PR's writer and toggle would sit on top of it unchanged. I am happy to rebase onto this branch, or to fold my config layer into it and close mine — whichever the maintainer prefers.

@itskylechung — you have spent more time in this area than I have, so if you have a few minutes I would value your read on #407, particularly on two choices:

1. workspace-level config vs per-slide `meta`, and whether they should be one mechanism
2. the preset list (`16:9`, `4:3`, `4:5`, `1:1`, `9:16`) alongside explicit pixel dimensions — is offering both worth it, or is one enough?

No rush and no expectation. Mostly I wanted you to know it exists rather than discover a duplicate later.

PR #198, opened 2026-06-03
fix: reject unsafe Windows asset names

chaochaoweb3:
## Summary
- reject Windows reserved device names such as `CON.png`, `AUX.txt`, and `COM1.svg` during asset filename validation
- reject filenames with leading/trailing whitespace so unsafe Windows trailing dots/spaces are not silently normalized
- add regression coverage for cross-platform asset name validation

## Verification
- RED: `corepack pnpm test -- --run packages/core/src/files/assets.test.ts` failed with `expected 'CON.png' to be null`
- GREEN: `corepack pnpm exec vitest run packages/core/src/files/assets.test.ts` passed, 9 tests
- `corepack pnpm test` passed, 16 files / 257 tests
- `corepack pnpm check` passed with one existing warning in `packages/core/src/http/request-guard.ts`
- `corepack pnpm --filter @open-slide/core typecheck` passed
- `corepack pnpm --filter @open-slide/core build` passed
- `git diff --check` passed

Note: root `corepack pnpm typecheck` failed in this local environment because Turbo could not find the package manager binary (`cannot find binary path`), so I ran the affected package typecheck directly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened asset filename validation to reject Windows reserved device basenames (including common COM/LPT variants) and names with leading/trailing whitespace.
  * Also disallows improperly formatted filenames such as those ending with a trailing dot, reducing Windows-specific file creation issues.
* **Tests**
  * Added coverage for the updated Windows reserved name and whitespace/trailing-dot validation rules.

### DIFF
--- a/.changeset/tidy-kids-march.md
+++ b/.changeset/tidy-kids-march.md
@@ -1,0 +1,5 @@
+---
+"@open-slide/core": patch
+---
+
+Tighten asset filename validation for Windows reserved names.
--- a/packages/core/src/files/assets.test.ts
+++ b/packages/core/src/files/assets.test.ts
@@ -37,9 +37,12 @@ describe('validateAssetName', () => {
     expect(validateAssetName('CON.png')).toBeNull();
     expect(validateAssetName('aux.txt')).toBeNull();
     expect(validateAssetName('COM1.svg')).toBeNull();
+    expect(validateAssetName('COM¹.svg')).toBeNull();
+    expect(validateAssetName('LPT².txt')).toBeNull();
     expect(validateAssetName('LPT9.jpg')).toBeNull();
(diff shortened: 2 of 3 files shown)


chaochaoweb3:
Addressed the CodeRabbit feedback in 9bed159: added a patch changeset for @open-slide/core, covered leading whitespace in the regression test, and extended Windows reserved COM/LPT matching to superscript digits.\n\nValidation run:\n- corepack pnpm exec vitest run packages/core/src/files/assets.test.ts\n- corepack pnpm --filter @open-slide/core typecheck\n- corepack pnpm check (passes with the existing request-guard optional-chain warning)\n- git diff --check

Issue #215, opened 2026-06-09
[Bug]: PDF export force-loads every registered @font-face, hanging/crashing the tab with CJK webfonts


poterpan:
### Summary

Exporting a deck to PDF pins the CPU and crashes the tab when the deck uses a CJK webfont (e.g. `Noto Sans TC` from Google Fonts). `waitForFonts()` in `packages/core/src/app/lib/print-ready.ts` calls `.load()` on every registered `FontFace` before printing. `FontFace.load()` **ignores `unicode-range`**, so it force-downloads *all* subset files of a CJK family — `Noto Sans TC:wght@300;400;500;700;900` registers **525 `@font-face` rules** (5 weights × ~105 unicode-range subsets) — including subsets the slides never use, and decodes them all at once. A system-font / Latin deck is unaffected because `document.fonts` stays small.

### Steps to reproduce

1. Scaffold a project and create a deck whose pages render Chinese text.
2. Load a CJK Google Font, e.g. `@import url('https://fonts.googleapis.com/css2?family=Noto+Sans+TC:wght@300;400;500;700;900&display=swap')`.
3. Add ~12–14 pages of Chinese content.
4. Trigger **Export → PDF** in the slide view.
5. The tab's CPU saturates, then the tab crashes before the print dialog is usable.

For contrast: the same deck with a system / Latin font exports normally.

### Expected behaviour

PDF export loads only the font subsets actually used by the rendered text (the browser's normal `unicode-range` lazy behaviour), completes quickly, and doesn't crash.

### Actual behaviour

`waitForFonts()` force-`.load()`s every registered face regardless of `unicode-range`, fetching and decoding all ~525 CJK subsets at once → 100% CPU, then tab crash. Relevant code:

```ts
// packages/core/src/app/lib/print-ready.ts
for (const face of document.fonts) {
  if (face.status !== 'loaded') pending.push(face.load());
}
await Promise.all(pending.map((p) => p.catch(() => undefined)));
```

### Affected package

`@open-slide/core`

### Version

`@open-slide/core@1.10.0`

### Environment

Node v22.19.0, pnpm 11.5.0, macOS 26.4 (build 25E246), Chrome 149.0.7827.102.

### Additional context

[...]

PR #216, opened 2026-06-09
feat(core): add webfont loading guidance to slide-authoring skill

poterpan:
Adds a **Webfonts** subsection to the `slide-authoring` skill.

The skill currently says "System stack unless the user specifies" but gives no guidance on *how* to load a webfont when one is actually needed. This documents the safe pattern:

- Load the stylesheet **once in `<head>`**, not inside a per-page component — every page mounts live at the same time (thumbnail rail, overview grid, PDF print root), so a per-page `@import` / `<link>` registers the whole `@font-face` set once *per page*.
- List only the weights actually used.
- Subset CJK families with `&text=` when the deck's text is fixed.

Authoring hygiene only — keeps the existing system-font-first default, no behaviour change. The second changed file is the CLI template copy, regenerated via `sync:template-skills`.

Pairs with #215, which reports the underlying `waitForFonts()` over-fetch that turns this per-page pattern into a tab crash on PDF export with CJK webfonts.

Note: `apps/web/content/docs/skills/slide-authoring.mdx` is a curated mirror of this skill — I left it untouched here, but happy to reflect the same guidance on the docs site if you'd like.

## Summary by CodeRabbit

* **Documentation**
  * Added webfonts guidance to slide-authoring: prefer system fonts when possible, load webfont stylesheets once in the page head with deduplication safeguards, only request needed font weights, and use CJK subsetting (e.g., Google Fonts text parameter) to reduce font sizes and speed up PDF/export times.

### DIFF
--- a/.changeset/webfont-skill-guidance.md
+++ b/.changeset/webfont-skill-guidance.md
@@ -1,0 +1,6 @@
+---
+"@open-slide/core": patch
+"@open-slide/cli": patch
+---
+
+Add webfont guidance to the slide-authoring skill: load the stylesheet once in `<head>`, list only the weights used, and subset CJK with `&text=`.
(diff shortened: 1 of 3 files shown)


1weiho (approved):

1weiho:
@poterpan Hi 👋🏻

PR #218, opened 2026-06-10
fix(core): guard Escape blur against IME composition in notes-drawer

### LINKED ISSUE #214: Notes textarea should not blur on Escape during IME composition
## Summary

When editing speaker notes with a Chinese/Japanese/Korean IME, pressing `Escape` to cancel the current IME composition causes the notes textarea to lose focus.

For IME users, `Escape` is commonly used to cancel candidate/composition input. Open Slide currently handles `Escape` as an editor shortcut and blurs the textarea, which interrupts typing.

## Environment

- Package: `@open-slide/core`
- Version observed: `1.10.0`
- Browser context: slide editor speaker notes drawer

## Current behavior

In `src/app/components/notes-drawer.tsx`, the notes textarea handles `Escape` like this:

```tsx
onKeyDown={(e) => {
  if (e.key === "Escape") {
    e.preventDefault();
    textareaRef.current?.blur();
  }
}}
```

So during Chinese IME input, pressing `Escape` to cancel the composition also triggers `blur()`.

## Expected behavior

When the user is composing text with an IME, `Escape` should be ignored by editor shortcuts. It should be left to the IME/browser composition flow.

After composition is finished, `Escape` can keep its current behavior of blurring or leaving the note editor.

## Suggested fix

Make the Escape handler composition-aware. A robust version usually checks both `nativeEvent.isComposing` and an explicit composition ref because browser behavior can vary around composition boundaries.

Example:

```tsx
const composingRef = useRef(false);

<textarea
  onCompositionStart={() => {
    composingRef.current = true;
  }}
  onCompositionEnd={() => {
    composingRef.current = false;
  }}
  onKeyDown={(e) => {
    if (e.key === "Escape") {
      if (e.nativeEvent.isComposing || composingRef.current) {
        e.stopPropagation();
        return;
      }

      e.preventDefault();
      textareaRef.current?.blur();
    } else if ((e.metaKey || e.ctrlKey) && e.key === "Enter") {
      e.preventDefault();
      void flush();
    }
  }}
/>
```

This is the same class of issue often described as an IME composition / keyboard shortcut conflict.

## Why this matters

This affects normal text entry for CJK users. The user intent is to cancel IME input, not to leave the notes textarea. Because speaker notes are long-form text, losing focus during composition is especially disruptive.

adawang1210:
## Problem
When editing speaker notes with a CJK IME (Bopomofo, Pinyin, etc.), pressing Escape to dismiss the candidate list also blurs the notes textarea, interrupting long-form typing. The Escape branch in the notes drawer's `onKeyDown` runs unconditionally during composition.

## Change
Add an early return in the Escape branch when `e.nativeEvent.isComposing` is true, so Escape is left to the IME during composition and keeps its blur behavior otherwise. Same pattern as the inspector content textarea fix in 7e00047.

## Testing
- `pnpm check`, `pnpm typecheck`, `pnpm test` all pass.
- Verified in `apps/demo` against the running dev server by dispatching `keydown` events on the notes textarea: Escape with `isComposing: true` keeps focus; Escape with `isComposing: false` still blurs.

## Related
Fixes #214

### DIFF
--- a/.changeset/ime-escape-notes-drawer.md
+++ b/.changeset/ime-escape-notes-drawer.md
@@ -1,0 +1,5 @@
+---
+'@open-slide/core': patch
+---
+
+Keep the notes textarea focused when Escape is pressed during IME composition.
--- a/packages/core/src/app/components/notes-drawer.tsx
+++ b/packages/core/src/app/components/notes-drawer.tsx
@@ -97,6 +97,9 @@ export function NotesDrawer({ slideId, index, total, initial }: Props) {
               }}
               onKeyDown={(e) => {
                 if (e.key === 'Escape') {
+                  // Escape during IME composition dismisses the candidate
+                  // list; it must not blur the textarea.
+                  if (e.nativeEvent.isComposing) return;
                   e.preventDefault();
                   textareaRef.current?.blur();
                 } else if ((e.metaKey || e.ctrlKey) && e.key === 'Enter') {


1weiho:
@adawang1210 Hi 👋🏻

1weiho (approved):

PR #221, opened 2026-06-10
fix(core): remove touchcancel listener on cleanup in useTouchSwipe

adawang1210:
## Problem
`useTouchSwipe` registers a `touchcancel` listener with an inline anonymous function, but the effect cleanup only removes `touchstart` and `touchend`. The effect's dependencies include `onPrev`/`onNext`, which are recreated on every page navigation in the Player, so each navigation re-runs the effect and stacks one more permanent `touchcancel` listener on the root element for the lifetime of the presentation.

## Change
Name the handler, register it, and remove it in the cleanup alongside the other two listeners.

## Testing
`pnpm check`, `pnpm typecheck`, and `pnpm test` all pass. No unit test added — the fix is listener bookkeeping inside a DOM-bound hook.

## Related
Found during an audit of `packages/core/src` event-listener cleanup.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an issue where touch event listeners were accumulating in presentation mode, improving performance during slide navigation.

### DIFF
--- a/.changeset/touch-swipe-cancel-listener-leak.md
+++ b/.changeset/touch-swipe-cancel-listener-leak.md
@@ -1,0 +1,5 @@
+---
+'@open-slide/core': patch
+---
+
+Remove the touchcancel listener on cleanup in the present-mode swipe handler so listeners no longer accumulate while navigating.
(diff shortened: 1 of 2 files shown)


1weiho (approved):

1weiho:
@adawang1210 Hi 👋🏻

PR #238, opened 2026-06-15
fix(core): include inline background-image assets in HTML export

### LINKED ISSUE #223: Export as HTML misses assets referenced from inline backgroundImage styles
## Problem
`Export as HTML` can miss images referenced from inline styles, for example a shared logo rendered via React inline `backgroundImage: url(...)`. Other images rendered with `` are exported correctly.

In the exported HTML/zip, the page renders but the logo is missing.

## Reproduction pattern
A slide imports a global asset and uses it as an inline background image:

```tsx
import hxLogo from "@assets/logo_hx_new.CJKvG6sq.png";

<div
  role="img"
  style={{
    backgroundImage: `url(${hxLogo})`,
    backgroundPosition: "50% 50%",
    backgroundRepeat: "no-repeat",
    backgroundSize: "contain",
  }}
/>
```

Then use the UI action `Export as HTML`. The exported file/package does not include or rewrite that logo URL, while regular `` assets are included.

## Expected behavior
The HTML exporter should collect and rewrite same-origin asset URLs referenced inside inline `style` attributes, including `background-image: url(...)`.

## Likely cause
In `src/app/lib/export-html.ts`, asset discovery currently checks:

- HTML `src`/`href`/`srcset` attributes via `findHtmlAssetUrls`
- stylesheet `url(...)` references via `findCssAssetUrls`

But URLs inside rendered inline style attributes are not parsed. React serializes `backgroundImage` into markup like `style="background-image: url(...)"`; because `findHtmlAssetUrls` only scans `src`/`href`/`srcset`, the exporter never fetches or rewrites that asset.

## Suggested fix
Extend HTML asset discovery to parse `style="..."` attributes and collect `url(...)` references from inline CSS declarations, reusing the existing URL validation/rewrite path.

## Environment
- `@open-slide/core`: 1.10.0
- Observed with a global asset imported via `@assets/...` and rendered as inline `backgroundImage`.

adawang1210:
## Problem

Fixes #223. HTML export drops assets referenced through inline `style="background-image:url(...)"`. Only `src`/`href`/`srcset` attributes were scanned, so inline-style background images vanished from the exported file.

## Cause

`findHtmlAssetUrls` in `packages/core/src/app/lib/export-html.ts` matched only `src`/`href`/`srcset`. The CSS `url(...)` scanner ran exclusively over the bundled stylesheet, never over the rendered page HTML, so inline-style backgrounds were never collected and therefore never bundled or rewritten.

## Fix

- Scan inline `url(...)` references in the rendered page HTML.
- Tolerate the HTML-entity-encoded quotes the DOM serializer emits (e.g. `url(&quot;/assets/x.png&quot;)`), as well as plain and unquoted forms.
- Existing `rewriteUrls` (string replace on the bare URL) already rewrites these once collected, so no change was needed there.

## Tests

Added `export-html.test.ts` covering src/srcset collection, inline background-image with entity-encoded / plain / unquoted forms, and rejection of data URIs and non-asset URLs.

`pnpm test`, `pnpm core typecheck`, and `biome check` on the changed files all pass.

## Release Notes

* **Bug Fixes**
  * Fixed HTML export to properly bundle inline CSS `background-image` assets instead of dropping them during export.

* **Tests**
  * Added comprehensive test coverage for asset URL discovery in HTML exports, including inline styles and multiple URL formats.

### DIFF
--- a/.changeset/html-export-inline-background-image.md
+++ b/.changeset/html-export-inline-background-image.md
@@ -1,0 +1,5 @@
+---
+"@open-slide/core": patch
+---
+
+Include inline `style` background-image assets in HTML export so they are bundled instead of dropped.
--- a/packages/core/src/app/lib/export-html.test.ts
+++ b/packages/core/src/app/lib/export-html.test.ts
@@ -1,0 +1,47 @@
+import { afterAll, beforeAll, describe, expect, it } from 'vitest';
+
+const originalWindow = (globalThis as { window?: unknown }).window;
+
+beforeAll(() => {
+  (globalThis as { window?: unknown }).window = {
+    location: { href: 'http://localhost:5173/', origin: 'http://localhost:5173' },
+  };
+});
+
+afterAll(() => {
+  if (originalWindow === undefined) {
+    delete (globalThis as { window?: unknown }).window;
+  } else {
+    (globalThis as { window?: unknown }).window = originalWindow;
+  }
+});
+
+const { findHtmlAssetUrls } = await import('./export-html');
+
+describe('findHtmlAssetUrls', () => {
+  it('collects src and srcset assets', () => {
+    const html = '<img src="/assets/a.png" srcset="/assets/b.png 1x, /assets/c.png 2x">';
+    expect(findHtmlAssetUrls(html)).toEqual(
+      expect.arrayContaining(['/assets/a.png', '/assets/b.png', '/assets/c.png']),
+    );
+  });
+
+  it('collects inline background-image with entity-encoded quotes', () => {
+    const html = '<div style="background-image:url(&quot;/assets/bg.png&quot;)"></div>';
+    expect(findHtmlAssetUrls(html)).toContain('/assets/bg.png');
+  });
+
+  it('collects inline background-image with plain quotes and without quotes', () => {
+    const html =
+      '<div style="background:url(\'/assets/x.jpg\')"></div><div style="background-image:url(/assets/y.webp)"></div>';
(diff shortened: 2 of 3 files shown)


mahsumaktas (commented):
Independently reviewed this (verified the diff and read the resulting `export-html.ts`). The fix is correct: `findInlineStyleAssetUrls` feeds the discovered `url(...)` references into the existing `urls` set, so they flow through fetch → `assets` map → `rewriteUrls` like any ``. Since the map key is the clean URL (a substring of the entity-encoded form), the `split/join` rewrite patches the inline style too. The `` / `srcset` path is untouched, and the `new Set([...])` dedup makes the extra return values harmless. #223 looks resolved.

Two optional observations, neither a blocker:

1. `findInlineStyleAssetUrls` actually scans the **entire** serialized HTML for `url(...)`, not just `style="…"` attributes. In practice `looksLikeAsset` (same-origin + asset extension) keeps this safe, but a deck that literally renders the text `url(/assets/x.png)` (say, in a `<code>` sample) would get that asset collected and rewritten too. Scoping the scan to `style="…"` / `style='…'` first would match the function name and tighten that edge. Current behavior is fine for real decks though.

2. The tests cover discovery well (entity-encoded, plain-quoted, unquoted, `data:`, non-asset rejection) but not the rewrite side — i.e. an inline-style URL actually being rewritten to `assets/…`. `rewriteUrls` is unchanged so the regression risk is low; an end-to-end case would just lock it in.

Thanks for the fix!

PR #239, opened 2026-06-15
fix(core): keep speaker notes aligned when deleting/duplicating a page

### LINKED ISSUE #222: Speaker notes can become misaligned or lost after copying/reordering pages
## Problem
When copying a slide page or moving pages to a new position, speaker notes can become attached to the wrong page. In some cases the notes appear to be lost.

This is serious during live presentations because the presenter notes no longer match the visible page.

## Observed behavior
- Copy a page that already has speaker notes.
- Or move a page before/after another page.
- The visible page order changes, but notes may stay bound to the old array index.
- A note that belonged to page A can show up on page B, or disappear after the operation.

## Expected behavior
- When a page is copied, its speaker notes should be copied with it.
- When a page is moved, its speaker notes should move with the page.
- Page content and notes should be treated as one logical unit during page structure operations.

## Likely cause
Speaker notes appear to be indexed by page position, while page copy/reorder operations mutate the `export default [...]` page array independently from the notes array. After insertion, deletion, duplication, or reorder, positional binding becomes stale.

## Suggested fix
Prefer binding notes to a stable page id instead of array index. If notes remain stored as an array, all page copy/reorder/delete/insert operations should update `notes` atomically with `export default [...]`, preserving order and length invariants.

## Environment
- `@open-slide/core`: 1.10.0
- Project uses `export const notes = [...]` alongside `export default [...] satisfies Page[]`.

adawang1210:
## Problem

Fixes #222. Speaker notes get misaligned or lost after a page is duplicated or deleted. Reordering pages handled notes correctly, but delete/duplicate did not.

## Cause

In `packages/core/src/vite/routes/slides.ts`, the reorder handler keeps the `notes` export in sync via `reorderNotesArrayInSource`, but the delete/duplicate handler only edited the default page array (`removePageFromDefaultExportInSource` / `duplicatePageInDefaultExportInSource`). The index-aligned `notes` array was left untouched, so every note after the affected index shifted by one.

## Fix

- Add `removeNotesElementInSource` and `duplicateNotesElementInSource` in `slide-ops.ts`, mirroring the existing `reorderNotesArrayInSource` (shared rebuild helper, same trailing-`undefined` trimming and shape guards).
- Apply them in the slides route delete/duplicate handler, returning a 422 if the `notes` export has an unexpected shape (matching the reorder path).

Both helpers no-op when the affected index falls past the recorded notes, and preserve template-literal notes verbatim.

## Tests

Added unit tests for both helpers (alignment after delete/duplicate, template-literal preservation, out-of-range no-ops, collapse to `[]`, and null on bad shapes). `pnpm test` (157 editing tests), `pnpm core typecheck`, and `biome check` on the changed files all pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Speaker notes now remain synchronized with pages when pages are deleted or duplicated.

### DIFF
--- a/.changeset/sync-notes-on-page-duplicate-delete.md
+++ b/.changeset/sync-notes-on-page-duplicate-delete.md
@@ -1,0 +1,5 @@
+---
+"@open-slide/core": patch
+---
+
+Keep speaker notes aligned with pages when deleting or duplicating a page.
--- a/packages/core/src/editing/slide-ops.test.ts
+++ b/packages/core/src/editing/slide-ops.test.ts
@@ -3,8 +3,10 @@ import os from 'node:os';
 import path from 'node:path';
 import { describe, expect, it } from 'vitest';
 import {
+  duplicateNotesElementInSource,
   duplicatePageInDefaultExportInSource,
   duplicateSlideDir,
+  removeNotesElementInSource,
   removePageFromDefaultExportInSource,
   reorderDefaultExportPagesInSource,
   reorderNotesArrayInSource,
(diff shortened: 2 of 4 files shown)


1weiho (approved):

1weiho:
@adawang1210 Hi 👋🏻
Nice fix! Thanks for the contribution, LGTM

PR #240, opened 2026-06-15
fix(core): show assets stored in nested folders in the asset library

### LINKED ISSUE #224: Asset library does not show files in nested slide asset folders
## Problem
The asset library view only shows files directly under `slides/<slideId>/assets/`, but it does not show files inside nested folders such as `slides/<slideId>/assets/keynote/`.

This is confusing because slide authoring guidance allows slide-specific images/videos/fonts under `slides/<id>/assets/`, and decks often organize screenshots into subfolders.

## Observed behavior
For a slide with this structure:

```
slides/ai-office-practice/assets/
  microsoft-ai-ceo-retraction.png
  terence-tao-ai-keynote.png
  keynote/
    keynote-01-2850x1591.png
    keynote-02-1672x941.png
    ...
```

Opening `/s/ai-office-practice?view=assets` only lists the top-level files. The files under `assets/keynote/` are not visible.

The dev asset endpoint confirms this behavior: `GET /__assets/ai-office-practice` returns only top-level files.

## Expected behavior
The asset library should either:

1. Recursively list assets under nested folders, preserving relative paths, or
2. Clearly document/enforce that the asset library only supports flat asset directories.

Given that authors commonly group screenshots, recursive listing seems more useful.

## Likely cause
The asset route uses non-recursive `fs.readdir(scopedDir)` and `validateAssetName(name)` only accepts a filename, not a relative path containing `/`. The UI also builds imports as `./assets/${asset.name}`, which assumes a flat directory.

## Environment
- `@open-slide/core`: 1.10.0

adawang1210:
## Problem

Fixes #224. Files placed in subfolders of a slide's `assets/` (or the global assets dir) never appear in the asset library — only top-level files are listed.

## Cause

In `packages/core/src/vite/routes/assets.ts`, the list handler read the scoped assets directory with a single non-recursive `fs.readdir`, and `validateAssetName` rejected any name containing a path separator. Nested files were therefore never enumerated, never served, and never matched for usage detection.

## Fix

- Add `listAssetFilesRecursive`, which walks the scoped directory and returns forward-slash relative paths.
- Add `validateAssetPath` (nested-aware sibling of `validateAssetName`): validates every segment, requires an extension only on the final segment, and rejects `..`, absolute, and backslash paths. The existing `path.resolve(...) + startsWith(dir + sep)` containment check remains the traversal boundary.
- Support nested paths end to end: list, serve, usages, upload (creates parent dirs), rename (keeps the asset in its directory), and delete. URLs keep the nested name in a single path segment via `encodeURIComponent` (`/` → `%2F`), so routing is unchanged.

## Tests

Added unit tests for `validateAssetPath` (nested accept, traversal/absolute/backslash/hidden-segment reject) and `listAssetFilesRecursive` (nested enumeration, missing dir). `pnpm test`, `pnpm core typecheck`, and `biome check` on the changed files all pass.

## Summary by CodeRabbit

* **New Features**
  * The asset library now supports and displays assets organized in nested folders. Assets can be accessed using paths like `logos/brand.svg`. Upload and rename operations preserve nested folder structures, enabling improved asset organization and discoverability.

### DIFF
--- a/.changeset/asset-library-nested-folders.md
+++ b/.changeset/asset-library-nested-folders.md
@@ -1,0 +1,5 @@
+---
+"@open-slide/core": patch
+---
+
+Show and serve assets stored in nested folders inside the asset library.
--- a/packages/core/src/files/assets.test.ts
+++ b/packages/core/src/files/assets.test.ts
@@ -1,5 +1,13 @@
+import fs from 'node:fs/promises';
+import os from 'node:os';
+import path from 'node:path';
 import { describe, expect, it } from 'vitest';
-import { mimeForFilename, validateAssetName } from './assets.ts';
+import {
+  listAssetFilesRecursive,
+  mimeForFilename,
+  validateAssetName,
+  validateAssetPath,
+} from './assets.ts';
 
 describe('validateAssetName', () => {
   it('accepts simple filenames with extensions', () => {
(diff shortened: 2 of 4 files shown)


mahsumaktas (commented):
Independently reviewed this (verified the diff + read the changed `files/assets.ts` and `vite/routes/assets.ts`). The backend is solid:

- **The path-traversal guard is airtight.** `validateAssetPath` rejects `\`, leading/trailing `/`, and per-segment `..`, leading `.` / `~`, control/forbidden chars, and empty segments, while `resolveAssetFile` / `resolveScopedAssetFile` still enforce `path.resolve(...) + startsWith(dir + sep)` containment. Defense-in-depth is preserved, and the new tests cover the traversal cases.
- The `%2F`-encoded single-segment URL keeps the serve regex working, since `new URL().pathname` doesn't decode `%2F`. Recursive listing returns clean forward-slash relative paths. Nicely done.

One should-fix and one nit:

1. **(should-fix) Renaming a nested asset from the UI fails.** The PR touches the backend + tests but not `app/components/asset-view.tsx`. `RenameCard` seeds its input with the full `asset.name` (e.g. `logos/brand.svg`); editing that to `logos/new.svg` posts a name containing a slash, which the rename handler's `validateAssetName` (single-segment) rejects with a 400. To make it work the user has to manually delete the `logos/` prefix and type just `new.svg`, with no hint in the UI. Might be worth seeding `RenameCard` with the bare filename (or splitting the UI side into a follow-up) — as-is, the "rename keeps the asset in its directory" behavior doesn't quite reach the user end-to-end.

2. **(nit) TOCTOU in the listing loop.** The per-entry `fs.stat` will throw and 500 the whole listing if a file is removed between enumeration and stat, and the recursive walk widens that window. A `try/catch` that skips on `ENOENT` (as CodeRabbit also noted) would harden it.

Thanks — the security side is handled well.

PR #217, opened 2026-06-09
fix(core): add third fill style to PPTX theme to avoid repair prompt

ridemountainpig:
## Summary

PPTX export was producing a theme whose `fmtScheme` `fillStyleLst` had only two entries. The OOXML schema requires exactly three fill styles, so PowerPoint showed a repair prompt when opening exported decks. This adds the missing third `gradFill` entry so the theme validates and opens cleanly.

Fix:
Microsoft PowerPoint 2026-06-09 20 47 19

## Changes

- `packages/core/src/app/lib/export-pptx.ts` — add the schema-required third fill style to the generated theme XML.

## Test plan

- Export a deck to PPTX and confirm PowerPoint opens it without a repair prompt.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected PPTX export theme generation to include required fill style entries, preventing PowerPoint from displaying repair prompts when opening exported presentations.

PR #243, opened 2026-06-16
fix(core): include speaker notes in PPTX export

### LINKED ISSUE #225: Image PPTX export opens with repair warning and omits speaker notes
## Problem
The current `Export as image PPTX` output opens in Microsoft PowerPoint with a repair warning. Speaker notes from `export const notes = [...]` are also not included in the generated PPTX.

PowerPoint warning observed on macOS:

> PowerPoint found a problem with content in ai-office-practice.pptx. PowerPoint can attempt to repair the presentation.

## Observed behavior
- Export a deck using the UI action `Export as image PPTX`.
- Open the downloaded `.pptx` in PowerPoint.
- PowerPoint asks to repair the file.
- After opening/repairing, speaker notes are missing.

## Expected behavior
- The exported image PPTX should open without repair.
- Speaker notes should be exported as PowerPoint notes, preserving the index-aligned `slide.notes` content.

## Likely causes
`src/app/lib/export-pptx.ts` hand-writes OOXML. The generated package currently:

- renders slides as full-page PNGs, but does not create `ppt/notesSlides/*` or `ppt/notesMasters/*`;
- does not add notes relationships from each `ppt/slides/_rels/slideN.xml.rels`;
- does not include notes content type overrides;
- is missing standard package parts that PowerPoint adds during repair, such as `docProps/*`, `ppt/viewProps.xml`, and `ppt/tableStyles.xml`.

## Local validation
A local patch that passes `slide.notes` into the image PPTX builder and emits `notesMaster`, per-slide `notesSlide`, notes relationships, content type overrides, `docProps`, `viewProps`, and `tableStyles` produced a PPTX with:

- valid ZIP structure (`ZipFile.testzip() == None`);
- XML parse success for all `.xml` files;
- one `ppt/notesSlides/notesSlideN.xml` per slide;
- each slide relationship file containing a `notesSlide` relationship;
- expected note text present in `notesSlideN.xml`.

## Environment
- `@open-slide/core`: 1.10.0
- PowerPoint for macOS

adawang1210:
## Problem

Fixes #225. Exporting to PPTX silently drops speaker notes — the generated deck has no notes slides at all.

## Cause

`buildImagePptx` in `packages/core/src/app/lib/export-pptx.ts` wrote one image slide per page but never produced any `notesSlide` parts, nor the notes master / content-type / relationship wiring they require. `slide.notes` was never read.

## Fix

For each page that has a (non-whitespace) note:

- Emit `ppt/notesSlides/notesSlide{idx}.xml` with the note text, splitting on newlines into paragraphs and XML-escaping `& < > "`.
- Link the slide to it via a `notesSlide` relationship in `slide{idx}.xml.rels`.
- Add a notes master (`notesMaster1.xml` + its own `theme2.xml`), referenced from `presentation.xml` (`notesMasterIdLst`, in schema order) and `presentation.xml.rels`.
- Register every new part in `[Content_Types].xml`.

When no page has notes, nothing changes — no notes master or notes slides are added, so decks without notes are byte-for-byte the same as before.

## Tests

Added `export-pptx.test.ts` (unzips the generated package via fflate): notes slides appear only for pages with notes, the slide↔notes relationship is wired, the notes master / theme2 / content-type overrides / `notesMasterIdLst` are present, XML-special characters are escaped, multi-line notes split into paragraphs, and whitespace-only notes are treated as empty.

`pnpm test` (full suite, 265 tests), `pnpm core typecheck`, and `biome check` on the changed files all pass.

## Summary by CodeRabbit

* **New Features**
  * Speaker notes are now exported with PowerPoint PPTX files and visible in the presentation's notes pane.

### DIFF
--- a/.changeset/pptx-export-speaker-notes.md
+++ b/.changeset/pptx-export-speaker-notes.md
@@ -1,0 +1,5 @@
+---
+"@open-slide/core": patch
+---
+
+Include speaker notes in the image PPTX export so they show up in PowerPoint's notes pane.
--- a/packages/core/src/app/lib/export-pptx.test.ts
+++ b/packages/core/src/app/lib/export-pptx.test.ts
@@ -1,0 +1,58 @@
+import { strToU8, unzipSync } from 'fflate';
+import { describe, expect, it } from 'vitest';
+import { buildImagePptx } from './export-pptx';
+
+const fakePng = strToU8('not-a-real-png');
+
+function unzipText(zip: Uint8Array): Record<string, string> {
+  const out: Record<string, string> = {};
+  const dec = new TextDecoder();
+  for (const [name, bytes] of Object.entries(unzipSync(zip))) {
+    out[name] = dec.decode(bytes);
+  }
+  return out;
+}
+
+describe('buildImagePptx', () => {
+  it('omits all notes parts when there are no notes', async () => {
+    const files = unzipText(await buildImagePptx([fakePng, fakePng]));
+    expect(Object.keys(files).some((p) => p.startsWith('ppt/notesSlides/'))).toBe(false);
+    expect(files['ppt/notesMasters/notesMaster1.xml']).toBeUndefined();
+    expect(files['ppt/presentation.xml']).not.toContain('notesMasterIdLst');
+  });
+
+  it('emits a notes slide only for pages that have notes', async () => {
+    const files = unzipText(
+      await buildImagePptx([fakePng, fakePng, fakePng], ['first note', undefined, 'third note']),
+    );
+
+    expect(files['ppt/notesSlides/notesSlide1.xml']).toContain('first note');
+    expect(files['ppt/notesSlides/notesSlide2.xml']).toBeUndefined();
+    expect(files['ppt/notesSlides/notesSlide3.xml']).toContain('third note');
+
+    // The slide links to its notes slide; one without notes does not.
+    expect(files['ppt/slides/_rels/slide1.xml.rels']).toContain('notesSlide1.xml');
+    expect(files['ppt/slides/_rels/slide2.xml.rels']).not.toContain('notesSlide');
+
+    // Notes master + theme2 and content-type overrides are wired in.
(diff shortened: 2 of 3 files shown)


mahsumaktas (commented):
Independently reviewed this (verified the OOXML wiring in `export-pptx.ts` + the tests). The notes plumbing is complete and internally consistent: the notesSlide part + per-slide `.rels` (→ slide, → notesMaster), `notesMaster1.xml` (+ rels → theme2), the slide→notesSlide relationship, the content-type overrides, and `notesMasterIdLst` placed in schema order in `presentation.xml` with `rId{n+3}` matching `presentation.xml.rels` exactly. No dangling relationships. The "byte-identical when there are no notes" guarantee genuinely holds — the gate is trim-aware and nothing notes-related is emitted when there are no notes — and the tests verify that.

On the repair-warning half of #225: as far as I can tell it's **already fixed on main** by #217. `themeXml` now carries the schema-required three `fillStyle` entries (CHANGELOG: "…ensuring the fill style list has the schema-required three entries"), and the report was on 1.10.0, which predates that. So the remaining real bug is the missing notes, which this addresses. (Caveat: I can't open the file in PowerPoint here — this is from static OOXML + the changelog. The package is structurally valid: every part has a content type, no dangling rels.)

One minor point:

1. `notesSlideXml` splits on `\n` without normalizing CRLF, so a `\r\n` note leaves a trailing `\r` in each `<a:t>`. CodeRabbit flagged this too; a `note.replace(/\r\n?/g, '\n')` before splitting handles it. Not a blocker.

For what it's worth, if the choice is between this and #226: this looks like the safer pick. It's gated so notes-free decks stay byte-identical (#226 adds a notesSlide to every page, changing their output), it ships tests, and it avoids #226's non-deterministic `new Date()` in `docProps/core.xml` and its hardcoded `lang="zh-CN"`. The docProps/viewProps/tableStyles #226 adds don't appear necessary for the repair prompt (that was the theme, fixed in #217). #226's one edge is a more "complete" notes page via a slide-image placeholder.

Thanks for gating it cleanly!

Issue #269, opened 2026-06-29
[Bug]: Non-ASCII slide id is listed but rejected by folder/slide mutations (can't move to a folder)


Yacolate0519-cmd:
## Summary

A slide whose directory name contains non-ASCII characters (e.g. CJK such as `推薦系統`) is discovered and shown in the sidebar, but it cannot be moved into a folder. The drag/assign action fails with `400 invalid slideId`. Discovery accepts the id while every mutation route rejects it, so the slide is visible but unusable for folder assignment — and the same guard also blocks rename, duplicate, delete, and asset listing for that slide.

## Steps to reproduce

1. `npx @open-slide/cli init repro && cd repro && pnpm install`
2. Create a slide whose **folder name is non-ASCII**, e.g. `slides/推薦系統/index.tsx` with a minimal default export.
3. `pnpm dev`, open http://localhost:5173.
4. The slide `推薦系統` appears in the sidebar (discovery accepts it).
5. Create a folder, then drag the slide into it (or trigger the assign action).
6. The request `PUT /__folders/assign` returns `400 { "error": "invalid slideId" }`; the slide is **not** added to the folder.

## Expected behaviour

Discovery and mutations should agree on what a valid slide id is. Either:
- a non-ASCII-id slide can be assigned to a folder like any other slide; **or**
- if non-ASCII ids are unsupported by design (the docs say `slides/<kebab-case-id>/`), the slide should not be silently listed as a normal, assignable slide — surface a clear dev-time warning ("rename to a kebab-case id") instead of a silent `400`.

## Actual behaviour

The slide is listed and looks normal, but assigning it to a folder silently fails with `400 invalid slideId`. Nothing in the UI explains why. The same `SLIDE_ID_RE` guard also rejects rename (`PATCH /__slides/:id`), duplicate, delete, and asset listing for that slide.

### Root cause (from source on `main`)

The scanner is permissive but the operations are ASCII-only:

- **Discovery does not constrain ids.** `findSlides` globs `*/index.{tsx,jsx,ts,js}` and `toId` takes the directory name verbatim — `packages/core/src/vite/open-slide-plugin.ts`. So any directory (including non-ASCII) becomes a `slideId` in `virtual:open-slide/slides`.
[...]

PR #271, opened 2026-06-29
fix(core): skip slide folders with non-ASCII ids during discovery

Yacolate0519-cmd:
## Problem

A slide whose directory name contains non-ASCII characters (e.g. a CJK folder name like `推薦系統`) shows up in the sidebar but can't be moved into a folder — the assign request fails with `400 invalid slideId`. The same guard also blocks rename, duplicate, delete, and asset listing for that slide. Full write-up and reproduction in #269.

## Root cause

Discovery and the mutation routes disagree on what a valid slide id is:

- `findSlides` globs `slides/*/index.*` and `toId` uses the directory name verbatim (`packages/core/src/vite/open-slide-plugin.ts`), so any folder — including a non-ASCII one — becomes a `slideId` in `virtual:open-slide/slides`.
- The editing/folder routes validate ids against `SLIDE_ID_RE = /^[a-z0-9_-]+$/i` (`packages/core/src/editing/slide-ops.ts`), enforced at e.g. `vite/routes/folders.ts` (assign), `vite/routes/slides.ts`, `vite/routes/assets.ts`.

So the slide is discoverable but every operation on it is rejected.

## Change (option A from #269)

Make discovery use the same rule as the routes:

- Filter out folders whose id fails `SLIDE_ID_RE` when generating `virtual:open-slide/slides`, so the sidebar no longer lists slides that can't be operated on.
- Emit a one-time `this.warn` per ignored folder, pointing the author to rename it to a kebab-case id.

A slide's display title (`meta.title`) is unaffected and can still be non-ASCII — only the folder **id** must be ASCII, which matches the documented `slides/<kebab-case-id>/` contract.

This is the conservative option A from #269. Happy to switch to first-class non-ASCII id support (option B) instead if you'd prefer — flagging the direction here rather than assuming.

## Testing

- New unit tests in `packages/core/src/vite/open-slide-plugin.test.ts` cover the all-ASCII case (nothing ignored) and the mixed case (non-ASCII folder excluded + reported).
- `pnpm format:check`, `pnpm lint`, `pnpm typecheck`, and `pnpm test` (275 passing) are all green locally.

Refs #269

## Summary by CodeRabbit

* **Bug Fixes**
  * Slide folders with non-ASCII-safe IDs are now excluded from the slide browser instead of causing later failures.
  * Added a single warning per ignored slide folder to reduce repeated noise.
* **Tests**
  * Added coverage for slide listing behavior with valid and invalid folder IDs.

### DIFF
--- a/.changeset/skip-non-ascii-slide-ids.md
+++ b/.changeset/skip-non-ascii-slide-ids.md
@@ -1,0 +1,5 @@
+---
+'@open-slide/core': patch
+---
+
+Skip slide folders whose id isn't ASCII-safe (e.g. CJK folder names) instead of listing slides that then fail to move into folders or be edited; warn once per ignored folder.
--- a/packages/core/src/vite/open-slide-plugin.test.ts
+++ b/packages/core/src/vite/open-slide-plugin.test.ts
@@ -1,0 +1,50 @@
+import fs from 'node:fs/promises';
+import os from 'node:os';
+import path from 'node:path';
+import { describe, expect, it } from 'vitest';
+import { generateSlidesModule } from './open-slide-plugin.ts';
+
+async function withSlidesRoot<T>(fn: (root: string) => Promise<T>): Promise<T> {
+  const root = await fs.mkdtemp(path.join(os.tmpdir(), 'open-slide-test-'));
+  try {
+    return await fn(root);
+  } finally {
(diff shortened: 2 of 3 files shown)


1weiho (approved):

1weiho:
@Yacolate0519-cmd Hi 👋🏻

PR #272, opened 2026-06-29
Fix editing/export correctness, dev-server hardening, and landing a11y

FurkanAksoyy:
## Summary

A batch of focused, independently-verified fixes across `packages/core`, `packages/cli`, and `apps/web` — correctness bugs in the editing/export engine, dev-server hardening, a Windows scaffolding bug, and a few landing-page accessibility/metadata cleanups.

All green: full test suite (276 tests), `packages/core` and `packages/cli` `tsc --noEmit`, `apps/web` `next typegen && tsc --noEmit`, and Biome lint.

## Core — editing engine

- **AST-based `meta.title` editing.** `updateMetaTitleInSource` now finds the `title` property via a Babel AST walk and rewrites only its value span. The old regex failed on a non-string value (e.g. `title: SOME_CONST`) and fell through to an inject path, producing an object with **two `title` keys**.
- **Reject overlapping splices.** `applyEdit` / `applyRevertAsset` apply splices by reverse-`from` string slicing; overlapping ranges corrupt each other and can still parse, slipping past the final guard. They are now rejected up front (half-open ranges, so touching boundaries are fine).
- **No duplicate asset imports.** Asset/placeholder ops in one edit now plan imports against accumulating state, so two ops on the same path (or colliding identifiers) reuse a single import instead of emitting a duplicate binding that fails the whole edit. (+ regression test)
- **CRLF import removal.** Removing an import line now consumes a trailing `\r\n`, not just `\n`, so Windows-authored files don't keep a stray blank line.

## Core — dev server & export

- **Prototype pollution fix.** `mergeDesign` skips `__proto__` / `constructor` / `prototype` keys, so a crafted `PUT /__design` body can no longer pollute `Object.prototype`. (+ regression tests)
- **Windows note focus-steal fix.** The notes HMR-suppression map is keyed by the Vite-normalized path; previously the backslash path never matched `ctx.file`'s forward slashes on Windows, so every note write remounted the tree and stole textarea focus mid-typing.
- **Asset upload hang fix.** An oversized upload called `req.destroy()`, which emits `'close'` (not `'end'`/`'error'`), so the body promise never settled and the request hung forever. It now resolves on `'close'` and returns 413; interrupted uploads are rejected.
- **Atomic manifest writes.** `writeManifest` serializes per-file writes and writes atomically (temp file + rename), so concurrent mutations can't interleave or truncate `.folders.json`.
- **Bounded note index** to keep a bogus huge index from allocating a giant padding array (OOM).
- **Full-bleed PPTX as a picture shape.** Each slide image is emitted as a slide-filling `<p:pic>` at the exact slide EMU extents with `stretch`/`fillRect`, instead of a `<p:bg>` background fill. The captured PNG and the frame are both exactly 16:9, so it maps 1:1 with no distortion, and a picture shape imports far more reliably into Google Slides (which handles per-slide backgrounds inconsistently). `noChangeAspect="0"` avoids letterboxing.
- **Export progress** counts only renderable pages, so a falsy page hole no longer leaves the bar short of 100%.
- **Emoji icon length** is measured by grapheme (`Intl.Segmenter`) instead of `String.length`, so ZWJ/flag/skin-tone emoji aren't falsely rejected by the 8-unit limit.

## CLI

- **`CLAUDE.md` scaffolding on Windows.** `cp()` copies the template's `CLAUDE.md`, which on a Windows checkout (`core.symlinks=false`) is the symlink's text placeholder `AGENTS.md` rather than a real file. The existence-gated repair skipped it, leaving the scaffolded project's `CLAUDE.md` containing the literal string `AGENTS.md`. It is now re-materialized unconditionally.

## Web (`apps/web`)

- **`prefers-reduced-motion`.** The `Anatomy` auto-cycle is gated on `useReducedMotion`, and a media block stops the looping marquee/caret animations and collapses the rest to their end state.
- **Keyboard-focusable slide player.** The inline player wrapper had an `onKeyDown` handler but no way to receive focus; it now gets `tabIndex`/`role` so arrow/space/page navigation works.
- **Theme toggle a11y.** Implements the ARIA radiogroup roving-tabindex pattern with arrow/Home/End navigation.
- **Slash-command highlighting** only triggers at the start of a line or after whitespace, so the `/cli` in `@open-slide/cli` isn't styled as a command.
- **Async sitemap mtimes.** Replaces top-level `execFileSync` (run at module evaluation) with promisified `execFile`, fetching per-page git mtimes concurrently.
- **De-duplicated metadata.** Drops the manually-declared favicon and `og:image` that duplicated the tags emitted by the `app/favicon.ico` and `app/opengraph-image.png` file conventions; adds `opengraph-image.alt.txt` to preserve the image alt text.

## Testing

`pnpm test` → 276 passing. New regression tests cover the design-merge prototype-pollution guard and the single-import-per-path behaviour.

## Summary by CodeRabbit

* **Accessibility**
  * Enhanced reduced-motion support and disabled landing animations when requested.
  * Improved keyboard navigation for theme selection and inline slide controls.
* **New Features**
  * Improved Open Graph and favicon generation behavior with updated alt text.
* **Bug Fixes**
  * Refined slash-command highlighting, slide metadata title updates, and slide export rendering.
  * Improved sitemap timestamps, edit/revert conflict detection, and shared asset import reuse.
  * Increased reliability for asset uploads, manifest updates, folder/slide assignment updates, and capped note index handling.
* **Security**
  * Hardened design merging against prototype pollution.
* **Tests**
  * Added coverage for splice conflict handling and merge security.

### DIFF
--- a/.changeset/fix-editing-export-devserver.md
+++ b/.changeset/fix-editing-export-devserver.md
@@ -1,0 +1,15 @@
+---
+"@open-slide/core": patch
+"@open-slide/cli": patch
+---
+
+Fix editing/export correctness and harden the dev server:
+
+- Edit meta.title via a Babel AST walk so a non-string title value no longer produces a duplicate `title` key.
+- Reject overlapping/nested splices in applyEdit/applyRevertAsset, and reuse a single import per asset path within an edit.
+- Prevent prototype pollution in the design merge (`PUT /__design`).
+- Fix Windows note-edit focus loss (Vite-normalized HMR path) and a hung asset upload on oversized/aborted bodies; bound the note index.
+- Serialize and atomically write the folders manifest.
+- Export full-bleed slides as a picture shape sized to the exact slide extents for reliable, undistorted Google Slides import.
+- Count emoji icon length by grapheme, and count only renderable pages for export progress.
+- CLI: always materialize `CLAUDE.md` from `AGENTS.md` so a Windows checkout no longer scaffolds a placeholder file.
--- a/apps/web/app/(home)/landing.css
+++ b/apps/web/app/(home)/landing.css
@@ -253,3 +253,15 @@ html.light .os-landing .vercel-oss-badge {
   outline-offset: 3px;
   border-radius: 6px;
 }
+
+/* Respect prefers-reduced-motion: stop the looping marquee/caret animations and
+   collapse the rest to their end state instead of running keyframes. */
+@media (prefers-reduced-motion: reduce) {
+  .os-landing *,
+  .os-landing *::before,
+  .os-landing *::after {
+    animation-duration: 0.01ms !important;
+    animation-iteration-count: 1 !important;
+    transition-duration: 0.01ms !important;
+  }
+}
--- a/apps/web/app/layout.tsx
+++ b/apps/web/app/layout.tsx
@@ -67,14 +67,9 @@ export const metadata: Metadata = {
     url: siteUrl,
     siteName: appName,
     locale: 'en_US',
-    images: [
-      {
-        url: '/opengraph-image.png',
-        width: 1200,
-        height: 630,
-        alt: `${appName} — React-first slide framework for AI agents`,
-      },
-    ],
+    // og:image (and its dimensions) are emitted automatically from the
+    // app/opengraph-image.png file convention; alt comes from the adjacent
+    // opengraph-image.alt.txt. Declaring images here too would duplicate the tag.
   },
   twitter: {
     card: 'summary_large_image',
@@ -94,10 +89,8 @@ export const metadata: Metadata = {
       'max-video-preview': -1,
     },
   },
-  icons: {
-    icon: '/favicon.ico',
-    shortcut: '/favicon.ico',
-  },
+  // favicon link is emitted automatically from the app/favicon.ico file
+  // convention; declaring icons here too would duplicate the <link rel="icon">.
 };
 
 export const viewport: Viewport = {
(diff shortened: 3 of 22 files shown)


FurkanAksoyy:
Thanks for the review — addressed all four findings:

1. **Splice overlap guard (`edit-ops`/`revert-asset`).** Extracted `splicesConflict`, which handles half-open ranges and zero-width inserts correctly, and now check **every** pair rather than only adjacent ones, so a zero-width insert nested inside a replacement can't slip through. Added unit tests. (68adb11)
2. **Manifest lost-update (`folders.ts`).** Added `updateManifest(file, mutator)` that serializes the whole read → mutate → write per file, and routed every folder/slide manifest mutation through it, so concurrent handlers can no longer clobber each other. (3d33b9a)
3. **Export progress/metadata (`export-pptx`).** The render loop now runs off a `renderablePages` list, so each slide's `SlidePageProvider` receives a compact index and the renderable-page total instead of the sparse array's length/indexes. (010e8fa)
4. **Missing changeset.** Added a patch changeset for `@open-slide/core` and `@open-slide/cli`. (4512a44)

All green: 280 tests, `packages/core` + `packages/cli` `tsc --noEmit`, `apps/web` `next typegen && tsc --noEmit`, and Biome lint.

FurkanAksoyy:
Also addressed the outside-diff finding on `assets.ts` (oversized upload):

`req.destroy()` could tear down the socket before the 413 was written. Now the handler stops buffering once over the limit (memory stays bounded) but keeps draining the body to its end, then sends a clean 413. (0518f83)

Still green: 280 tests, core/cli/web typecheck, Biome lint.

PR #313, opened 2026-07-14
feat(core): improve shared element transition (magic move) support

1weiho:
## Summary

Deep-dive into the shared element transition (magic move) implementation, driven by the hacks a real 190-page keynote deck had to make. Two framework fixes plus full skill documentation:

- **Measure shared-element rects before the enter/exit phases start** ([slide-transition-layer.tsx](https://github.com/1weiho/open-slide/blob/shared-element-magic-move/packages/core/src/app/components/slide-transition-layer.tsx)). Phase keyframes apply immediately (`fill: 'both'`), so a `translateY` in the enter's first keyframe used to offset every measured target rect — clones landed a few px off and snapped at morph end, forcing authors into pure-opacity transitions on every morphing page. Measuring first removes that constraint; `sharedElements` now composes with any transition family.
- **Add `unstable_useIsActivePage()`** ([step-context.tsx](https://github.com/1weiho/open-slide/blob/shared-element-magic-move/packages/core/src/app/lib/step-context.tsx)). The runtime mounts a fresh instance of the outgoing page to snapshot its exit state, and the dev UI mounts every page for thumbnails/overview/presenter/print — decks had to parse `?p=` from the URL to keep typewriters, connector draws, and audio from replaying there (wrong in thumbnails of the current page, the presenter window, and print). The hook reads the `isActivePage` flag `StepHost` already threads through; hostless mounts default to `false` so they render the settled final state.
- **Document shared element transitions in the `slide-authoring` skill** (core copy + CLI template copy kept in sync): opt-in contract, FLIP semantics (id pairing, in-place fades for unmatched ids, bidirectional morphs), seven authoring rules (deterministic geometry, no transform on the shared node, single-child style merge, morph-timed reveals, hold pages…), and taste guidance on when to reach for a magic move.

## Changesets

- `@open-slide/core` **minor** — new public API `unstable_useIsActivePage()`
- `@open-slide/core` **patch** — measurement-order fix
- `@open-slide/core` + `@open-slide/cli` **patch** — skill docs

## Testing

- `pnpm typecheck` and `pnpm test` (305 tests) pass; biome clean on changed files.
- The measurement reorder stays within the same synchronous layout effect (all animations still start in the same frame before paint), so existing opacity-only decks are unaffected.
- No WAAPI test harness exists for the transition layer; runtime verification is easiest by giving a morphing page a transform-bearing enter — before this fix the clones land offset and snap, after it they land flush.

## Summary by CodeRabbit

- **New Features**
  - Added `unstable_useIsActivePage()` to identify the live audience-facing page and control entrance animations.
  - Improved shared-element (“magic move”) transitions by measuring elements before enter and exit animations begin.

- **Documentation**
  - Added guidance and review requirements for configuring shared-element transitions, including matching IDs, stable geometry, and animation gating.

### DIFF
--- a/.changeset/shared-element-measure-first.md
+++ b/.changeset/shared-element-measure-first.md
@@ -1,0 +1,5 @@
+---
+"@open-slide/core": patch
+---
+
+Measure shared-element rects before the enter/exit phases start so transform keyframes no longer offset magic-move targets.
--- a/.changeset/shared-element-skill-docs.md
+++ b/.changeset/shared-element-skill-docs.md
@@ -1,0 +1,6 @@
+---
+"@open-slide/core": patch
+"@open-slide/cli": patch
+---
+
+Document shared element transitions (magic move) in the slide-authoring skill.
--- a/.changeset/use-is-active-page.md
+++ b/.changeset/use-is-active-page.md
@@ -1,0 +1,5 @@
+---
+"@open-slide/core": minor
+---
+
+Add unstable_useIsActivePage() so pages can gate entrance animations to the live audience-facing instance.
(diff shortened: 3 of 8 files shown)


1weiho:
@coderabbitai help me review, thanks!

PR #366, opened 2026-07-23
fix(core): accept Unicode/CJK slide ids for discovery and mutations

### LINKED ISSUE #269: [Bug]: Non-ASCII slide id is listed but rejected by folder/slide mutations (can't move to a folder)

stantheman0128:
## Summary

Fixes #269 (option B).

After #271, non-ASCII slide folders were dropped from discovery so they no longer appeared half-broken. This goes further: Unicode letter/digit slide ids (including CJK such as `推薦系統`) are first-class for both listing and mutations.

- Broaden `SLIDE_ID_RE` to `/^[\p{L}\p{N}_-]+$/u` (still rejects spaces, dots, separators; path-traversal checks unchanged)
- Discovery keeps the same filter, so CJK ids now list instead of being warned away
- Percent-encode slide id segments in client `/__slides` and `/__assets` fetches; decode path segments on the server
- Sibling paths covered: folder assign, rename, duplicate, delete, asset list

## Evidence (Windows 11, demo at `http://localhost:5173`, slide folder `slides/推薦系統/`)

```text
listedInSlidesModule: true
createFolder: { status: 200, id: "f-2e9222cb" }
assign: { status: 200, body: { ok: true } }
assignedInManifest: true
assetsList: { status: 200, count: 0 }
rename: { status: 200, body: { ok: true, slideId: "推薦系統", name: "推薦系統演示" } }
duplicate: { status: 200, slideId: "推薦系統-copy" }
deleteDuplicate: { status: 200 }
```

Unit tests:

```text
pnpm exec vitest run packages/core/src/editing/slide-ops.test.ts packages/core/src/vite/open-slide-plugin.test.ts
Test Files  2 passed (2)
Tests  67 passed (67)
```

## AI assistance

This change was prepared with AI assistance (Cursor/Grok). I reviewed the diff, ran the commands above, and verified the product path with a live CJK slide id on Windows.

## Test plan

- [x] Unit: CJK id accepted by `SLIDE_ID_RE`, listed by `generateSlidesModule`, duplicate/rm/resolve work
- [x] Live: assign / rename / duplicate / delete / asset list for `推薦系統`
- [ ] Maintainer: optional UI drag-into-folder smoke on a CJK-named slide

## Summary by CodeRabbit

* **New Features**
  * Slide folder IDs now support Unicode letters and digits, including CJK, for discovery and editing operations.
  * Asset and slide links now use safe encoding for IDs and asset names.
* **Bug Fixes**
  * Improved handling of percent-encoded URL path segments with safer decoding and clearer errors for invalid slide or asset identifiers.
  * Updated slide and asset routes to validate decoded IDs consistently and prevent malformed paths.

### DIFF
--- a/.changeset/cjk-slide-ids.md
+++ b/.changeset/cjk-slide-ids.md
@@ -1,0 +1,5 @@
+---
+'@open-slide/core': patch
+---
+
+Accept Unicode (including CJK) slide folder ids for discovery and mutations.
--- a/packages/core/src/app/lib/assets.ts
+++ b/packages/core/src/app/lib/assets.ts
@@ -15,9 +15,15 @@ export type AssetEntry = {
 
 export type UploadOptions = { overwrite?: boolean };
 
+function assetsScopeUrl(slideId: string, ...parts: string[]): string {
+  const segments = [encodeURIComponent(slideId), ...parts.map((p) => encodeURIComponent(p))];
+  return `/__assets/${segments.join('/')}`;
+}
+
 export async function listAssets(slideId: string): Promise<AssetEntry[]> {
-  const res = await fetch(`/__assets/${slideId}`);
-  if (!res.ok) throw new Error(`GET /__assets/${slideId} ${res.status}`);
+  const url = assetsScopeUrl(slideId);
+  const res = await fetch(url);
+  if (!res.ok) throw new Error(`GET ${url} ${res.status}`);
   const data = (await res.json()) as { assets?: AssetEntry[] };
   return data.assets ?? [];
 }
@@ -28,7 +34,7 @@ export async function uploadAsset(
   opts: UploadOptions = {},
 ): Promise<Response> {
   const qs = opts.overwrite ? '?overwrite=1' : '';
-  return fetch(`/__assets/${slideId}/${encodeURIComponent(file.name)}${qs}`, {
+  return fetch(`${assetsScopeUrl(slideId, file.name)}${qs}`, {
     method: 'POST',
     headers: {
       'content-type': file.type || 'application/octet-stream',
@@ -39,7 +45,7 @@ export async function uploadAsset(
 }
 
 async function renameAsset(slideId: string, from: string, to: string): Promise<Response> {
-  return fetch(`/__assets/${slideId}/${encodeURIComponent(from)}`, {
+  return fetch(assetsScopeUrl(slideId, from), {
     method: 'PATCH',
     headers: { 'content-type': 'application/json' },
     body: JSON.stringify({ name: to }),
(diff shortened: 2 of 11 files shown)


stantheman0128:
Addressed CodeRabbit feedback on `b6a8538`: asset filename segments now use `decodePathSegment` (fail-closed 400 on malformed percent-encoding), matching scope decoding. `pnpm exec vitest run packages/core/src/vite/open-slide-plugin.test.ts packages/core/src/editing/slide-ops.test.ts` — 67 passed.

Stan Shih (@stantheman0128)

stantheman0128:
Follow-up: fork already had the same fix at \2f20927\ before this session's local commit; no additional push needed.

PR #367, opened 2026-07-23
fix(core): select imported/shared components in the inspector

### LINKED ISSUE #327: [Bug]: Inspector can't select any element rendered by an imported/shared component
### Summary

In the dev-server inspector, you can only pick DOM elements that were authored as host JSX directly in `slides/<id>/index.tsx`. Any element rendered by a component imported from another file or package is unselectable — including all of its text. Decks built on a shared component library (which open-slide's own docs encourage over per-slide duplication) therefore lose element-picking for essentially all of their content.

### Steps to reproduce

```ts
// components/Heading.tsx  (a separate module — this is the key)
export const Heading = ({ children }: { children: React.ReactNode }) =>
  <div style={{ fontSize: 64 }}>{children}</div>;

// slides/demo/index.tsx
import { Heading } from '../../components/Heading';
export default [() => <Heading>Click me</Heading>];
```

- Click "Click me" in the inspector → not selectable.
- Inline the same markup (`<div style={{fontSize:64}}>Click me</div>`) directly in `index.tsx` → selectable.

The trigger is specifically that the host `<div>` is created in a different file. If Heading is defined in the same slide file it works, which masks the bug in simple decks.

### Expected behaviour

Clicking any rendered element in the inspector selects it and resolves to the JSX that produced it in `slides/<id>/index.tsx` — regardless of whether that JSX is an inline host tag or a component imported from another file or package. For a click on text rendered by a shared component, the inspector should select the text node and map it to the component's call-site line in the slide.

### Actual behaviour

Only elements authored as inline host JSX directly in the slide file are selectable. Any element rendered by a component imported from another file/package — including all of its text — is unselectable: the click produces no selection and no source mapping. In a deck built on a shared component library, this means effectively all content (every heading, body line, card, list) can't be picked, while stray inline `` / `<div>` tags still can — making it look like "text specifically is broken."

### Affected package

@open-slide/core

### Version

@open-slide/core@1.13.5

### Environment

Node 22.20.0, pnpm 11.9.0, macOS 15.7.3 (24G419), Chrome 150.0.7871.124, @open-slide/core 1.13.5

### Additional context

(This is all AI-generated RCA)

**Root cause**
Two mechanisms both key off "was this DOM node born in the slide file", and a shared component fails both:

1. `data-slide-loc` injection — `src/vite/loc-tags-plugin.ts`
- `isSlideSourceFile()` only transforms files under the deck's own `slides/` dir, so library/component files are never tagged.
- `isTaggableJsxName()` only tags lowercase host tags or names in the hardcoded `FORWARDING_COMPONENTS = new Set(['ImagePlaceholder'])`.

2. Fiber fallback — `src/app/lib/inspector/fiber.ts`, `findSlideSource()`

-  Match requires `src.fileName` to end with `/slides/${slideId}/index.tsx` and `(!opts.hostOnly || isHost)`, where `isHost = fiber.stateNode instanceof HTMLElement`.
- Every inspector call site passes `{ hostOnly: true }`: `inspect-overlay.tsx:43,52,64` and `inspector-panel.tsx:65,1098`.

Walking up from a click on shared-component text: the host `<div>`'s `_debugSource` points at the library file (no match) → the next fiber up is the component call-site fiber (`<Heading>`, whose `_debugSource` is the slide file, a real match) → but it's a function component, not a host, so `hostOnly: true` rejects it → the walk reaches the root and returns `null`. Unselectable.

**Proposed fix (preferred)**
Allow the fallback to match the nearest component fiber whose `_debugSource` is the slide file, anchoring the highlight to the nearest descendant host — which `findSlideSource` already tracks in anchor. Concretely, drop `hostOnly` (or make it prefer-host-then-fall-back):

```ts
- const hit = findSlideSource(el, slideId, { hostOnly: true });
+ const hit = findSlideSource(el, slideId);
```

For the text-click case this resolves to the `<Heading>` call-site line and highlights the rendered `<div>` — exactly the JSX the author wrote. No deck changes required, and inline host JSX keeps working (it still matches first, lower in the walk).

**Alternative / complementary fix**
Make `FORWARDING_COMPONENTS` configurable via `OpenSlideConfig` so a component library can register its primitives; those components then spread the injected `data-slide-loc` onto their root host (as `ImagePlaceholder` already does). This preserves strict `hostOnly` semantics and gives precise per-element anchoring, at the cost of each registered component forwarding the attribute.

stantheman0128:
## Summary

Fixes #327.

In inspect mode, clicking an element rendered by an imported/shared component produced no selection (or, when nested under a tagged slide host, selected the wrong ancestor). Host JSX authored outside `slides/` never gets `data-slide-loc`, and the fiber fallback required a host fiber whose `_debugSource` pointed at the slide file, so component call sites were skipped.

- Prefer an exact `data-slide-loc` on the clicked element
- Then walk fibers and allow slide-file component call sites (drop `hostOnly` at inspector call sites), anchoring highlight to the nearest descendant host
- Only then fall back to a tagged ancestor (HMR-immune last resort)
- Covers nested shared components and text inside shared hosts

## Evidence (product path)

Scratch e2e fixture `shared-select` with `components/shared.tsx` (`Heading` / `Card`), Playwright against the live inspector.

**Before (main HEAD):** click Shared heading click target with Inspect on -> inspector panel stays closed (`panelVisible: false`).

**After (this branch):** same click opens the panel with Element text `Shared heading click target` (host styles 64px / weight 700). Nested Card > Heading (`Nested shared heading`) also selects.

```text
pnpm exec vitest run packages/core/src/app/lib/inspector/fiber.test.ts
Test Files  1 passed (1)
Tests  13 passed (13)

pnpm exec playwright test e2e/tests/inspector.spec.ts -g "shared component|nested shared|selecting an element opens"
  ok selecting an element opens the panel with its tag and text
  ok selecting text inside an imported shared component opens the panel
  ok selecting text inside a nested shared component opens the panel
```

Screenshots / logs live under `handoff/evidence-open-slide-327/` on the contributor machine (before unselectable, after shared + nested selected).

## What was not tested

- Style edits that require the shared component to forward `style` / props (selection + source mapping are fixed; forwarding is still per-component)
- Configurable `FORWARDING_COMPONENTS` (issue alternative); not needed for selection

## AI assistance

This change was prepared with AI assistance (Cursor/Grok). I reviewed the diff, ran the unit and Playwright checks above, and verified the inspector click path before/after on the shared-component fixture.

## Test plan

- [x] Unit: shared call site, nested call site, tagged-ancestor must not steal, hostOnly still rejects when set
- [x] E2E: select text inside imported Heading; select nested Heading under Card; inline host regression
- [ ] Maintainer: quick click-through in a deck that imports a local component library

## Summary by CodeRabbit

* **New Features**
  * The inspector can now select elements rendered by imported or shared components, including nested shared components.
  * Selected shared component content appears in the inspector’s “Element text” field.

* **Bug Fixes**
  * Improved target resolution when clicked elements lack direct location metadata, including line and column fallback behavior.

* **Tests**
  * Added end-to-end and unit coverage for shared component selection and nested behavior.

### DIFF
--- a/.changeset/shared-inspector-select.md
+++ b/.changeset/shared-inspector-select.md
@@ -1,0 +1,5 @@
+---
+'@open-slide/core': patch
+---
+
+Make the inspector selectable for elements rendered by imported/shared components.
--- a/packages/core/e2e/fixture/components/shared.tsx
+++ b/packages/core/e2e/fixture/components/shared.tsx
@@ -1,0 +1,9 @@
+import type { ReactNode } from 'react';
+
+export function Heading({ children }: { children: ReactNode }) {
+  return <div style={{ fontSize: 64, fontWeight: 700 }}>{children}</div>;
+}
+
+export function Card({ children }: { children: ReactNode }) {
+  return <div style={{ padding: 24, border: '1px solid #444', borderRadius: 12 }}>{children}</div>;
+}
--- a/packages/core/e2e/fixture/slides/shared-select/index.tsx
+++ b/packages/core/e2e/fixture/slides/shared-select/index.tsx
@@ -1,0 +1,18 @@
+﻿import type { Page, SlideMeta } from '@open-slide/core';
+import { Card, Heading } from '../../components/shared';
+
+export const meta: SlideMeta = {
+  title: 'Shared Select',
+  createdAt: '2026-01-01T00:00:00.000Z',
+};
+
+const Only: Page = () => (
+  <>
+    <Heading>Shared heading click target</Heading>
+    <Card>
+      <Heading>Nested shared heading</Heading>
+    </Card>
+  </>
+);
+
+export default [Only] satisfies Page[];
--- a/packages/core/e2e/fixture/tsconfig.json
+++ b/packages/core/e2e/fixture/tsconfig.json
@@ -13,5 +13,5 @@
     "skipLibCheck": true,
     "types": ["@open-slide/core/env"]
   },
-  "include": ["slides/**/*", "open-slide.config.ts"]
+  "include": ["slides/**/*", "components/**/*", "open-slide.config.ts"]
 }
--- a/packages/core/e2e/tests/inspector.spec.ts
+++ b/packages/core/e2e/tests/inspector.spec.ts
@@ -140,4 +140,24 @@ test.describe('inspector editing', () => {
     await editorCanvas(page).getByText('Editable headline').click();
     await expect(page.locator('aside[data-inspector-ui]')).toBeVisible();
   });
+
+  test('selecting text inside an imported shared component opens the panel', async ({ page }) => {
+    await openSlide(page, 'shared-select');
+    await page.getByTitle('Inspect').click();
+    await editorCanvas(page).getByText('Shared heading click target').click();
+
+    const panel = page.locator('aside[data-inspector-ui]');
+    await expect(panel).toBeVisible();
+    await expect(panel.getByPlaceholder('Element text')).toHaveValue('Shared heading click target');
+  });
+
+  test('selecting text inside a nested shared component opens the panel', async ({ page }) => {
+    await openSlide(page, 'shared-select');
+    await page.getByTitle('Inspect').click();
+    await editorCanvas(page).getByText('Nested shared heading').click();
+
+    const panel = page.locator('aside[data-inspector-ui]');
+    await expect(panel).toBeVisible();
+    await expect(panel.getByPlaceholder('Element text')).toHaveValue('Nested shared heading');
+  });
 });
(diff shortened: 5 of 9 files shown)


stantheman0128:
Addressed CodeRabbit feedback on `6ab55d4`: `findElementByLine` fallback now requires both `line` and `column` before returning an anchor (avoids same-line sibling collisions). `pnpm exec vitest run packages/core/src/app/lib/inspector/fiber.test.ts` — 13 passed.

Stan Shih (@stantheman0128)

stantheman0128:
Follow-up: fork already had the column-aware fallback at \5f7678e\ (exact column first, line-only fallback); no additional push needed.

PR #382, opened 2026-07-29
feat(core): drag to move and corner handles to resize in the inspector

### LINKED ISSUE #380: [Feature]: Drag and resize the selected element directly on the canvas
## Problem

Inspect mode edits almost everything about the selected element: its text, size, weight, line height, tracking, alignment, colours, even the image behind it. The two properties it cannot touch are the two most visual ones, position and size.

So the smallest and most iterative kind of edit ("this heading sits 40px too low", "this card is slightly too wide") is the one that has to leave the canvas, either back to the agent or into `slides/<id>/index.tsx` to guess a number, reload, and look again. Every other adjustment in the panel is immediate and reversible. Layout is the one that is not.

## Proposed solution

In Inspect mode, dragging the body of the selected element moves it, and four corner handles resize it with the opposite edge anchored.

Both gestures end as ordinary `set-style` operations, the same ones the panel already emits:

- move: `translate`
- resize: `width` and `height`, plus `translate` when a top or left corner is dragged, so the anchored edge stays put

Save, undo, redo and the write-back into the `.tsx` then work with no new machinery.

Two details that were not obvious while building it:

- Movement uses the standalone `translate` property rather than `transform`, so it composes with whatever `transform` the slide already declares instead of overwriting it.
- The canvas scale has to be resolved from the selected element's own canvas, because the same slide is also rendered into the thumbnail rail and the overview grid at much smaller scales. Reading the wrong one multiplies every drag.

## Alternatives considered

- **Keep going through the agent.** Works, but a nudge becomes a round trip, and nudging is inherently iterative.
- **Edit the source by hand.** Requires knowing the target coordinate before seeing the result.
- **Arrow-key nudging.** A good complement, but it does not cover resize and is still blind for large moves.

## Which package would this affect?

`@open-slide/core (runtime, inspector, present mode, CLI)`

## Additional context

No new dependencies. About 300 lines in a new `drag-resize-layer.tsx`, plus a small hook-up in `inspect-overlay.tsx`.

I have this implemented and would be glad to open a PR. `format:check`, `lint`, `typecheck`, the unit suite and the inspector e2e suite all pass against it. Happy to change the interaction to whatever you would prefer before writing anything further.

bruno-begh:
Closes #380.

## Problem

Inspect mode edits everything about the selected element except the two most visual properties: where it sits and how big it is. Position and size are the only adjustments that have to leave the canvas, either back to the agent or into the `.tsx` to guess a number, reload, and look again.

## Change

Dragging the body of the selected element moves it. Four corner handles resize it, with the opposite edge anchored.

Both gestures end as ordinary `set-style` operations, the same ones the panel already emits, so Save, undo, redo and the write-back into the `.tsx` need no new machinery:

- move: `translate`
- resize: `width` and `height`, plus `translate` when a top or left corner is dragged

Three things worth a reviewer's attention:

- Movement uses the standalone `translate` property rather than `transform`, so it composes with whatever `transform` the slide already declares instead of overwriting it.
- The canvas scale is resolved from the element's own canvas via `closest('[data-osd-canvas]')`. The thumbnail rail and the overview grid render the same slide at much smaller scales, and reading the wrong one multiplies every drag.
- The DOM is restored to its pre-gesture state before `bufferOps` runs, because that call snapshots the current inline value for undo. Without the restore, undo would return to the dragged position rather than the original one.

A gesture ends with a click the overlay would otherwise read as "select whatever is under the pointer", so exactly one click is swallowed, and the listener disarms itself after 300ms if no click follows.

## Testing

- `pnpm format:check`, `pnpm lint`, `pnpm typecheck`: clean
- `pnpm test`: 305 passing
- `e2e/tests/inspector.spec.ts`: 7 passing
- By hand in `apps/demo`: drag plus each of the four handles, then Save, undo and redo, checking the values written into the `.tsx` and that an existing `transform` on the element survives.
- `packages/core/tools/verify-drag-resize.mjs` drives the gestures against the `verify-drag-resize` fixture and asserts the source was rewritten with the expected geometry. Run it with the demo dev server up.

No new dependencies.

The fixture slide and the verification script are there to make the behaviour checkable in a clone. Happy to drop either if you would rather keep the demo free of them.

## Summary by CodeRabbit

* **New Features**
  * Added interactive drag-and-drop repositioning for selected elements in the inspector.
  * Added corner handles to resize selected elements, with updates reflected immediately and saved as style edits.
  * Enabled the interactive selected-frame overlay in Inspect mode.
* **Tests**
  * Added end-to-end verification covering drag, resize, and saving the resulting style changes.

### DIFF
--- a/.changeset/inspector-drag-resize.md
+++ b/.changeset/inspector-drag-resize.md
@@ -2,4 +2,4 @@
 '@open-slide/core': minor
 ---
 
-Drag the selected element to reposition it and use the corner handles to resize, saved back to source as style edits.
+Add drag-to-move and corner resize handles for the selected element in the inspector, saved back to source as style edits.
--- a/.changeset/inspector-drag-resize.md
+++ b/.changeset/inspector-drag-resize.md
@@ -1,0 +1,5 @@
+---
+'@open-slide/core': minor
+---
+
+Drag the selected element to reposition it and use the corner handles to resize, saved back to source as style edits.
--- a/apps/demo/slides/verify-drag-resize/index.tsx
+++ b/apps/demo/slides/verify-drag-resize/index.tsx
@@ -1,0 +1,31 @@
+import type { Page, SlideMeta } from '@open-slide/core';
+
+// Fixture for tools/verify-drag-resize.mjs. The heading text is what the script
+// clicks on, so keep it in sync with the script's default selector.
+
+const Cover: Page = () => (
+  <div
+    style={{
+      width: '100%',
+      height: '100%',
+      background: '#0d0d10',
+      color: '#f4f4f5',
+      display: 'flex',
+      flexDirection: 'column',
+      justifyContent: 'center',
+      padding: '0 160px',
+      gap: 32,
+    }}
+  >
+    <p style={{ fontSize: 28, letterSpacing: 6, color: '#8b5cf6', margin: 0 }}>DRAG AND RESIZE</p>
+    <h1 style={{ fontSize: 120, fontWeight: 700, lineHeight: 1.05, margin: 0 }}>
+      Original headline of the cover
+    </h1>
+    <p style={{ fontSize: 40, color: '#a1a1aa', margin: 0 }}>
+      This paragraph exists so the fixture has more than one text block.
+    </p>
+  </div>
+);
+
+export const meta: SlideMeta = { title: 'Verify drag and resize', createdAt: '2026-07-29' };
+export default [Cover] satisfies Page[];
--- a/packages/core/src/app/components/inspector/drag-resize-layer.tsx
+++ b/packages/core/src/app/components/inspector/drag-resize-layer.tsx
@@ -125,13 +125,29 @@ export function DragResizeLayer({
       suppressNextClick();
     };
 
+    // The browser can take the pointer away mid-gesture (a touch turning into a
+    // scroll, the window losing the device). No pointerup follows, so the drag
+    // has to be unwound here or the element keeps the position it happened to
+    // have when the gesture died.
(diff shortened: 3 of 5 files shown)


bruno-begh:
Addressed all three points, pushed in d6391db.

**`pointercancel` (major).** Correct, and the resize handle was the worse of the two: its `pointermove` and `pointerup` listeners were only detached inside `onUp`, so a cancelled gesture leaked them permanently and the element kept resizing on every later pointer move. Both gestures now detach and restore the pre-gesture inline styles on `pointercancel`.

**Resize seed.** Also correct. I measured it rather than assume: with `box-sizing: content-box` and `width:100px; padding:10px; border:5px`, `getBoundingClientRect().width` is 130 while `style.width` is `100px`. `getComputedStyle().width` returns `100px` under both box models, which is exactly what gets written back, so the seed now comes from there. That also drops the element-local transform concern, since the computed value is a layout value.

**Changeset wording.** Reworded, though not to the suggested phrasing. Every entry in `packages/core/CHANGELOG.md` opens with a plain verb (`Add`, `Fix`, `Restore`, `Stabilize`), so it now reads "Add drag-to-move and corner resize handles for the selected element in the inspector, saved back to source as style edits."

I left the docstring coverage warning alone. CONTRIBUTING asks contributors to "default to writing no comments" and to document only the non-obvious *why*, which is what the file does.

Re-verified after the change: `format:check`, `lint`, `typecheck`, the unit suite and `e2e/tests/inspector.spec.ts` all pass, and `tools/verify-drag-resize.mjs` still confirms both gestures persist the expected geometry to the `.tsx`.

pablospe:
any update on this?

PR #383, opened 2026-07-29
feat(core): resolve deck dependencies before serving

### LINKED ISSUE #381: [Feature]: Resolve a deck's fonts and assets before the dev server starts
## Problem

A deck is portable code, but what it renders with is not. The fonts it names and the assets it imports live outside the file, so a deck that looks right on the machine that wrote it can open elsewhere with the browser quietly substituting a different family. Nothing reports it. You find out when the layout looks slightly off, or later, in an export.

The more a deck travels, the more this costs: a theme naming a Google font, a colleague cloning the repo, a fresh laptop, a rebuild months later.

## Proposed solution

Resolve the gap at `open-slide dev`, before serving.

On start, read the decks and themes, collect the font families actually in use (including the runtime's own, which decks inherit without naming it anywhere), and compare against what the machine has installed. Whatever is missing gets installed:

1. from the matching npm package when there is one,
2. otherwise from the `google/fonts` repository, which is the stable public source of installable outlines. (`fonts.google.com/download` returns HTML, and the CSS API returns woff2, so neither of them installs.)

Installation is per user, so no administrator rights are involved, and it works on Windows, macOS and Linux.

Assets follow the same idea: a deck can ship `slides/<id>/assets.manifest.json` mapping file names to URLs, and anything missing on disk is fetched.

Also available on its own as `open-slide preflight`, with `--no-install` to only list what is missing.

## Alternatives considered

- **Document the required fonts in the deck's README.** Depends on discipline, and says nothing at the moment it matters.
- **Fail the boot when something is missing.** Hostile. A missing decorative font should not stop you from working.
- **Vendor every font into the repo.** Bloats the repository and duplicates the same files across decks.
- **Warn without installing.** Better than silence, but still leaves the work to the reader.

## Which package would this affect?

`@open-slide/core (runtime, inspector, present mode, CLI)`

## Additional context

Failure is never fatal: a network error or an unresolvable family becomes a line in the report, never an exception.

It does not run under CI, where the machine is discarded after the run and the check would only buy latency and one more way for a job to fail. `--no-preflight` turns it off anywhere else.

No new dependencies. I have this implemented and would be glad to open a PR. `format:check`, `lint`, `typecheck`, the unit suite and the CLI e2e suite all pass against it.

bruno-begh:
Closes #381.

## Problem

A deck is portable code, but what it renders with is not. The fonts it names and the assets it imports live outside the file, so a deck that looks right on the machine that wrote it opens elsewhere with the browser quietly substituting a different family. Nothing reports it.

## Change

`open-slide dev` now resolves that gap before serving. It reads the decks and themes, collects the font families in use (including the runtime's own, which decks inherit without naming it anywhere), and installs what the machine is missing, from the `google/fonts` repository. That is the stable public source of installable outlines: `fonts.google.com/download` returns HTML and the CSS API returns woff2, so neither yields something an operating system can install.

An outline the workspace already carries in `node_modules` is reused, but no package is ever installed to obtain one. Writing a dependency and a lockfile entry into someone's project as a side effect of starting a dev server would be the wrong trade, and font packages ship woff2 anyway.

Installation is per user, so no administrator rights are involved: `%LOCALAPPDATA%\Microsoft\Windows\Fonts` plus the `HKCU` registration on Windows, `~/Library/Fonts` on macOS, `~/.local/share/fonts` on Linux.

Assets follow the same idea: a deck can ship `slides/<id>/assets.manifest.json` mapping file names to URLs, and anything missing on disk is fetched.

Also exposed on its own as `open-slide preflight`, with `--no-install` to only report.

### On not getting in the way

- Nothing here can take the server down. A network failure or an unresolvable family becomes a line in the report, never an exception.
- It does not run under CI, where the machine is discarded after the run and the check would only buy latency and one more way for a job to fail. That also keeps it out of this repo's own e2e suite.
- `--no-preflight` turns it off anywhere else, mirroring the existing `--no-skills-check`.

Static faces are preferred over the variable build when a family ships both, because a static file names its weight in the file itself, which is what a system font install exposes.

## Testing

- `pnpm format:check`, `pnpm lint`, `pnpm typecheck`: clean
- `pnpm test`: 309 passing, including four new cases covering when the check is skipped
- `e2e/tests/cli.spec.ts` and `e2e/tests/dev-api.spec.ts`: 12 passing
- By hand on Windows: `open-slide preflight --no-install` reports across the 19 demo decks without touching the system, and leaves `package.json` byte-identical. Against a scratch workspace naming a family the machine did not have, `--no-install` reported it as pending and a full run downloaded and installed it where the OS then resolves it.

I have only exercised the install path on Windows. The macOS and Linux paths are the conventional per-user font directories, but they deserve a second pair of eyes from someone on those platforms.

No new dependencies.

## Summary by CodeRabbit

- **New Features**
  - Added a `preflight` command to verify required fonts and deck assets, with optional installation/download of missing items.
  - Development startup now runs preflight checks by default (with opt-out via `--no-preflight`).
  - Added `--no-install` to skip asset installation during preflight.
  - Includes system font discovery and per-user font installation support.
- **Bug Fixes**
  - Preflight failures no longer block starting the development server.
- **Tests**
  - Added coverage for preflight run/skipping behavior in CI and opt-out scenarios.

### DIFF
--- a/.changeset/dependency-preflight.md
+++ b/.changeset/dependency-preflight.md
@@ -1,0 +1,5 @@
+---
+'@open-slide/core': minor
+---
+
+Check and install the fonts and assets a deck needs on dev start, with an `open-slide preflight` command.
--- a/packages/core/src/cli/dev.ts
+++ b/packages/core/src/cli/dev.ts
@@ -5,11 +5,13 @@ import chalk from 'chalk';
 import { createServer, mergeConfig } from 'vite';
 import { createViteConfig } from '../vite/config.ts';
 import { DEV_SUPERVISED_ENV, RESTART_EXIT_CODE } from '../vite/routes/restart.ts';
+import { runPreflight, shouldRunPreflight } from './preflight.ts';
 
 export interface DevOptions {
   port?: number;
   host?: string | boolean;
   open?: boolean;
+  preflight?: boolean;
 }
 
 export async function dev(opts: DevOptions = {}): Promise<void> {
@@ -21,6 +23,16 @@ export async function dev(opts: DevOptions = {}): Promise<void> {
 }
 
 async function startServer(opts: DevOptions): Promise<void> {
+  if (shouldRunPreflight(opts)) {
+    try {
+      await runPreflight(process.cwd());
+    } catch (err) {
+      process.stderr.write(
+        `${chalk.yellow('preflight:')} dependency check failed (${(err as Error).message})\n`,
+      );
+    }
+  }
+
   const base = await createViteConfig({ userCwd: process.cwd() });
   const config = mergeConfig(base, {
     server: {
@@ -93,6 +105,7 @@ function supervise(opts: DevOptions): void {
 
 function devArgs(opts: DevOptions): string[] {
   const args = ['dev', '--no-skills-check'];
+  if (opts.preflight === false) args.push('--no-preflight');
   if (opts.port !== undefined) args.push('--port', String(opts.port));
   if (opts.host === true) args.push('--host');
   else if (typeof opts.host === 'string') args.push('--host', opts.host);
--- a/packages/core/src/cli/preflight.test.ts
+++ b/packages/core/src/cli/preflight.test.ts
@@ -1,0 +1,21 @@
+import { describe, expect, it } from 'vitest';
+import { shouldRunPreflight } from './preflight.ts';
+
+describe('shouldRunPreflight', () => {
+  it('runs by default on a developer machine', () => {
+    expect(shouldRunPreflight({}, {})).toBe(true);
+  });
+
+  it('is skipped by --no-preflight', () => {
+    expect(shouldRunPreflight({ preflight: false }, {})).toBe(false);
+  });
+
+  it('is skipped under CI', () => {
+    expect(shouldRunPreflight({}, { CI: 'true' })).toBe(false);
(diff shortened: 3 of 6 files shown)


bruno-begh:
Both points were right. Pushed in 1695d2d, and the PR description is updated to match.

**Installing a package into the user's project (major).** Removed the `npm install` path entirely rather than adding `--no-save`. The deeper problem is the one you named second: it spawned plain `npm` inside what may well be a pnpm or yarn workspace. And it rarely bought anything, because font packages target browsers and ship woff2, which no operating system can install. An outline the workspace already carries in `node_modules` is still reused; everything else now comes from `google/fonts`, which was already doing the real work. Net effect is 26 lines removed.

**Exception safety (major).** Wrapped the per-family loop. You were right that this contradicted the PR's own claim: `dev.ts` happened to catch, but `open-slide preflight` has no `try/catch` above it, so the "reported without stopping" promise only held for one of the two callers.

Verified after the change:

- `preflight` across the demo's 19 decks leaves `package.json` byte-identical.
- Against a scratch workspace naming a family the machine did not have, `--no-install` reported it as pending, and a full run downloaded it from `google/fonts` and installed it where the OS then resolves it.
- `format:check`, `lint`, `typecheck` and the 309 unit tests pass.

Docstring coverage warning left as is, per CONTRIBUTING's "default to writing no comments".

Issue #173, opened 2026-05-28
[Feature]: support subpath deployment (Vite base path & React Router basename)


dm-yeu:
### Problem

Currently, `open-slide` cannot be hosted under a subfolder path on a web server (e.g., `http://my-intranet/my-slides/` or GitHub Pages `http://username.github.io/repo/`). 

When deployed under a subpath, Nginx successfully serves the built index.html and assets (CSS/JS files are fetched with `200 OK`), but the browser displays a React Router `404 Not Found` page.

This happens because:

1. `createViteConfig` in `@open-slide/core` hardcodes `configFile: false`, preventing standard custom `vite.config.ts` base paths from taking effect.
2. In `@open-slide/core/src/app/app.tsx`, the `<BrowserRouter>` component is instantiated without a `basename` prop. It defaults to the root `/`, causing history routing to fail when matched against a subpath like `/my-slides/`.

### Proposed solution

1. **Configurable Base Path**: Add an optional `base?: string` option in `OpenSlideConfig` inside `open-slide.config.ts`:
   ```ts
   // open-slide.config.ts
   import type { OpenSlideConfig } from '@open-slide/core';
   const config: OpenSlideConfig = {
     base: '/my-slides/',
   };
   export default config;

2. **Vite base config**: Pass this value into the Vite configuration generated by `createViteConfig`:
   ```ts
   // @open-slide/core (vite config builder)
   return {
     base: config.base ?? '/',
     root: APP_ROOT,
     configFile: false,
     ...
   }
   ```

3. **Router Basename**: Automatically pass Vite's `import.meta.env.BASE_URL` to React Router's `<BrowserRouter>` basename prop so that the client-side router matches the subpath correctly:
   ```tsx
   // @open-slide/core/src/app/app.tsx
   export function App() {
     return (
       <BrowserRouter basename={import.meta.env.BASE_URL}>
         <Routes>
           ...
         </Routes>
       </BrowserRouter>
     );
   }
   ```

### Alternatives considered

An alternative would be exposing a configuration toggle to use `HashRouter` instead of `BrowserRouter`. Hash-routing is completely path-agnostic (using `#/s/slide-id`), which would allow the static build to run under any subpath—and even directly from the local file system (`file://` protocol)—without any pathing configurations.

### Which package would this affect?

@open-slide/core (runtime, inspector, present mode, CLI)

### Additional context

[...]

PR #393, opened 2026-08-05
fix: honor configured base for dev authoring APIs

### LINKED ISSUE #392: [Bug]: Dev authoring APIs ignore configured base under subpath hosting
## Summary

When `OpenSlideConfig.base` is a non-root path, the OpenSlide UI and Vite module graph can render correctly under that path, but the dev authoring APIs do not follow it.

Client code still issues root-absolute requests such as `fetch('/__folders')`, `fetch('/__notes')`, and `fetch('/__edit')`. Behind a reverse proxy that exposes OpenSlide only beneath a subpath, those requests escape the OpenSlide mount and hit the host origin instead.

Making the client URLs base-relative is necessary but not sufficient: the corresponding Connect middleware is registered only at root paths such as `server.middlewares.use('/__folders', ...)`. A request arriving as `/<base>/__folders` therefore misses the route and falls through to Vite's HTML handling.

This leaves folder mutations, inline editing, notes, comments, asset mutations, and server-management actions unavailable when OpenSlide is hosted beneath a subpath.

## Steps to reproduce

1. Configure OpenSlide with a non-root base:

   ```ts

   const config: OpenSlideConfig = {
     base: '/my-slides/',
     port: 5173,
   };

   export default config;
   ```

2. Run `open-slide dev` behind a reverse proxy that exposes the server only at `/my-slides/`.
3. Open `https://<host>/my-slides/`. The library, decks, modules, deep links, and HMR work.
4. Exercise an authoring action such as creating/assigning a folder, editing slide text, or saving speaker notes.
5. Observe that the browser requests root-origin paths such as `https://<host>/__folders` or `https://<host>/__edit`, outside `/my-slides/`, and receives a host 404.
6. Probe the dev server directly to isolate the server-side half:

   ```
   GET /__folders             -> 200 application/json
   GET /my-slides/__folders   -> 200 text/html (SPA fallback; route missed)
   ```

## Expected behaviour

When `base` is configured, all OpenSlide-owned dev API URLs remain beneath that base, and the dev server resolves those routes there. Authoring features should work through the same subpath as rendering, navigation, modules, and HMR.

## Actual behaviour

The application renders and HMR works beneath the configured base, but authoring API calls either:

- escape the mount because the client uses root-absolute `/__*` URLs; or
- miss OpenSlide's Connect middleware when sent beneath the configured base because the server routes are root-only.

The failures are generally handled by the UI, so rendering can appear healthy while mutations are inert.

## Affected package

`@open-slide/core`

## Version

`@open-slide/core@1.17.1`; the root-absolute client calls and root-only middleware registration are also present on current `main`.

## Environment

Linux container, Vite 5.4.21, Chromium, OpenSlide dev server behind a reverse proxy at a nested path.

## Additional context

This appears to be a follow-up gap in the subpath support completed by #173 and #207. That work covered Vite's base, React Router's basename, presenter URLs, and virtual imports, but not the dev authoring API path space.

Examples on current `main`:

- Client: [`packages/core/src/app/lib/folders.ts`](https://github.com/1weiho/open-slide/blob/main/packages/core/src/app/lib/folders.ts)
- Client: [`packages/core/src/app/lib/inspector/use-notes.ts`](https://github.com/1weiho/open-slide/blob/main/packages/core/src/app/lib/inspector/use-notes.ts)
- Client: [`packages/core/src/app/lib/inspector/use-editor.ts`](https://github.com/1weiho/open-slide/blob/main/packages/core/src/app/lib/inspector/use-editor.ts)
- Server: [`packages/core/src/vite/routes/folders.ts`](https://github.com/1weiho/open-slide/blob/main/packages/core/src/vite/routes/folders.ts)

A likely direction is to centralize client dev-API URL construction around `import.meta.env.BASE_URL` and make server route registration normalize or explicitly support the configured base, with coverage for both `base: '/'` and a nested base.

iddogino:
Fixes #392

When `OpenSlideConfig.base` is a non-root path, the UI renders fine beneath the base but all dev authoring APIs break: the client issues root-absolute `fetch('/__folders')`-style requests that escape the mount behind a reverse proxy, and the Connect middleware is only registered at root paths, so requests arriving as `/<base>/__*` fall through to Vite's SPA HTML handling.

## Changes

**Client** — new `devApiUrl()` helper in `app/lib/dev-api.ts` joins `import.meta.env.BASE_URL` (the same signal #173/#207 used for the router basename and presenter URLs) with the dev API path. Every `/__*` fetch across folders, slides, editing, notes, comments, design, assets, svgl, update, and restart now goes through it, as do the constructed asset/svgl proxy URLs used as image sources.

**Server** — new `mountDevRoute()` / `devRoutePaths()` helpers in `vite/routes/mount.ts` read the resolved `server.config.base` and mount each route beneath the base first, keeping the root mount for direct probes and pre-base clients (with `base: '/'` registration is unchanged — a single root mount). Used by all `routes/*` registrations plus the design and notes plugins. Asset URLs embedded in JSON responses are base-prefixed via `withBase()`.

## Tests (written first, TDD)

- `vite/routes/base-routing.test.ts` — HTTP-level reproduction of the issue over a connect-style mount stack: with `base: '/my-slides/'`, `GET /my-slides/__folders` previously fell through to `text/html` (the exact failure in the report) and now returns `application/json`; root probes and SPA fallback for non-API paths still behave. Plus a route-inventory sweep asserting every dev API route group (including `/__design` and `/__notes`) mounts both beneath a nested base and at root.
- `vite/routes/mount.test.ts` — unit coverage for `devRoutePaths`/`withBase`/`mountDevRoute` across `/`, nested, trailing-slash-less, deep, and relative bases.
- `app/lib/dev-api.test.ts` — `joinBase`/`devApiUrl` under `BASE_URL='/'` and `'/my-slides/'` (via `vi.stubEnv`), plus a source scan guarding against future root-absolute `fetch('/__…')` call sites in `src/app`.

Red run before the fix: the nested-base routing test failed with `expected 'text/html' to contain 'application/json'`; after the fix all 329 tests pass. `pnpm typecheck` and `pnpm check` are clean, and a changeset (patch, `@open-slide/core`) is included.

## Summary by CodeRabbit

- **Bug Fixes**
  - Development APIs now work correctly when the application is hosted under a custom or nested base path.
  - Asset URLs and development tools—including editing, comments, notes, designs, folders, slides, updates, and server controls—now resolve reliably in subpath deployments.

- **Tests**
  - Added coverage for root and nested base paths, route mounting, asset URLs, and development API requests.

### DIFF
--- a/.changeset/brave-paths-follow.md
+++ b/.changeset/brave-paths-follow.md
@@ -1,0 +1,5 @@
+---
+'@open-slide/core': patch
+---
+
+Keep dev authoring APIs (folders, editing, notes, comments, assets, server actions) under the configured `base` so they work behind subpath hosting.
--- a/packages/core/src/app/components/image-placeholder.tsx
+++ b/packages/core/src/app/components/image-placeholder.tsx
@@ -1,6 +1,7 @@
 import { type CSSProperties, type HTMLAttributes, useRef, useState } from 'react';
 import { toast } from 'sonner';
 import { uploadWithAutoRename } from '@/lib/assets';
+import { devApiUrl } from '@/lib/dev-api';
 import { useLocale } from '@/lib/use-locale';
 
 export type ImagePlaceholderProps = {
@@ -206,7 +207,7 @@ function pickImageFile(files: FileList): File | null {
 async function handleDrop(slideId: string, file: File, line: number, column: number) {
   const { ok, entry } = await uploadWithAutoRename(slideId, file);
   if (!ok || !entry) throw new Error('upload failed');
-  const res = await fetch('/__edit', {
+  const res = await fetch(devApiUrl('/__edit'), {
     method: 'POST',
     headers: { 'content-type': 'application/json' },
     body: JSON.stringify({
--- a/packages/core/src/app/components/sidebar/sidebar-footer.tsx
+++ b/packages/core/src/app/components/sidebar/sidebar-footer.tsx
@@ -4,6 +4,7 @@ import { useEffect, useState } from 'react';
 import { toast } from 'sonner';
 import { Button } from '@/components/ui/button';
 import { Tooltip, TooltipContent, TooltipProvider, TooltipTrigger } from '@/components/ui/tooltip';
+import { devApiUrl } from '@/lib/dev-api';
 import { format, useLocale } from '@/lib/use-locale';
 import { useRestartServer } from '@/lib/use-restart-server';
 
@@ -23,7 +24,7 @@ export function SidebarFooter() {
   useEffect(() => {
     if (!import.meta.env.DEV) return;
     let cancelled = false;
-    fetch('/__update-check')
+    fetch(devApiUrl('/__update-check'))
       .then((res) => (res.ok ? (res.json() as Promise<UpdateCheck>) : null))
       .then((data) => {
         if (!cancelled && data?.outdated) setUpdate(data);
@@ -43,7 +44,7 @@ export function SidebarFooter() {
     setUpdateStatus('running');
     setOpen(true);
     try {
-      const res = await fetch('/__update-package', { method: 'POST' });
+      const res = await fetch(devApiUrl('/__update-package'), { method: 'POST' });
       if (!res.ok) throw new Error('update failed');
       setUpdateStatus('done');
       toast.success(t.home.updatePackageDone);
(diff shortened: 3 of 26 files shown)


iddogino:
@1weiho would you mind taking a look?

PR #421, opened 2026-08-24
fix(core): check batch edit changed flag before clearing pending edits

### LINKED ISSUE #408: [Bug]: The batch edit route reports success but does not save the file
### Summary

The route /__edit/batch can process an edit with no error. But the edit does not always change the file.

Example response:

 ```
{"ok":true,"changed":false,"results":[{"ok":true}]}
 ```

This response is correct. The file did not change. The server did not write the file. But the app does not check the value of changed. The function commitEdits() in inspector-provider.tsx checks only the value of ok for each edit. It does not check the value of changed for the batch.

### Steps to reproduce

1. Open a deck in the Inspector.
2. Select a text element.
3. Change a style value. For example, change the font size.
4. Click Save. The save succeeds and writes the new value.
5. Select the same element again.
6. Set the style value to the same value as in step 4.
7. Click Save.
8. Check the network response for /__edit/batch. The response shows "changed":false.
9. Check the Save bar. The Save bar shows success, with no warning.

### Expected behaviour

The Save bar must not show success for an edit that did not change the file. The app must show a warning, in the same way as the single-edit route's NoOpEditError.

### Actual behaviour

The Save bar shows success. The app gives the user no warning. The user can lose data after a page refresh, with no sign that the save did not work.

### Affected package

@open-slide/core

### Version

@open-slide/core@1.18.0

### Environment

Node 24.19.0, npm 11.17.0, Debian 12 (bookworm), Firefox+Chrome

### Additional context

**Root cause (from source)**

 The single-edit route /__edit already checks this case. The file use-editor.ts has this code:

```js
if (body.changed === false) throw new NoOpEditError()
```

The batch route /__edit/batch has no matching check, in use-editor.ts (client) and inspector-provider.tsx (commitEdits()).

**Suggested fix**

 1. Change applyEdits() in use-editor.ts. Make it return the value of changed for the batch.
 2. Change commitEdits() in inspector-provider.tsx. Make it check this value.
 3. If changed is false, do not clear the pending edit. Show a message like the message for NoOpEditError.

fatima-n09:
## Description:
      Fixes #408
    
  ## Problem
      The batch edit route /__edit/batch reports success even when the
      file did not change. commitEdits() only checks per-edit ok flags
      but ignores the batch-level changed field. This causes the Save
      bar to show success for no-op edits, and users can lose data after
      a page refresh without any warning.
    
      The single-edit route /__edit already handles this case by throwing
      NoOpEditError when body.changed === false.
    
   ## Fix
      - Update applyEdits() in use-editor.ts to return the batch-level
        changed flag alongside the per-edit results
      - Update commitEdits() in inspector-provider.tsx to check the
        changed flag and show a no-op warning instead of clearing pending
        edits
      - Add noOpEdit translation key to all locale files (en, zh-tw,
        zh-cn, ja)
    
   ## Files changed
      - packages/core/src/app/lib/inspector/use-editor.ts
      - packages/core/src/app/components/inspector/inspector-provider.tsx
      - packages/core/src/locale/en.ts
      - packages/core/src/locale/types.ts
      - packages/core/src/locale/zh-tw.ts
      - packages/core/src/locale/zh-cn.ts
      - packages/core/src/locale/ja.ts

## Summary by CodeRabbit

* **Bug Fixes**
  * Inspector edits now clearly distinguish successful changes, failed edits, and requests that leave files unchanged.
  * Failed edit notifications are prioritized, while failed and unchanged edits remain pending for review.
  * Save confirmations now appear only when changes are successfully saved.
  * Added a notification when requested values already match the file.

* **Localization**
  * Added no-change edit messages in English, Japanese, Simplified Chinese, and Traditional Chinese.

### DIFF
--- a/.changeset/fix-batch-edit-noop.md
+++ b/.changeset/fix-batch-edit-noop.md
@@ -1,0 +1,5 @@
+---
+'@open-slide/core': patch
+---
+
+Check batch edit changed flag before clearing pending edits and show a no-op warning when the file did not change.
--- a/.changeset/fix-batch-edit-noop.md
+++ b/.changeset/fix-batch-edit-noop.md
@@ -2,4 +2,4 @@
 '@open-slide/core': patch
 ---
 
-Check batch edit changed flag before clearing pending edits and show a no-op warning when the file did not change.
+Warn when a batch edit does not change the file and keep the edits pending.
--- a/packages/core/src/app/components/inspector/inspector-provider.tsx
+++ b/packages/core/src/app/components/inspector/inspector-provider.tsx
@@ -241,7 +241,7 @@ type InspectorCtx = {
   // close) is what actually writes to disk; `cancelEdits` reverts.
   bufferOps: (line: number, column: number, anchor: HTMLElement, ops: EditOp[]) => void;
   pendingCount: number;
-  commitEdits: () => Promise<void>;
+  commitEdits: () => Promise<boolean>;
   cancelEdits: () => void;
   committing: boolean;
   openCrop: (anchor: HTMLImageElement) => void;
@@ -632,9 +632,9 @@ export function InspectorProvider({
     [applyOpsRaw, snapshotForOps, restoreSnapshot, findAnchor, history, ensureInstanceId],
   );
 
-  const commitEdits = useCallback(async () => {
+  const commitEdits = useCallback(async (): Promise<boolean> => {
     const buckets = pendingRef.current;
-    if (buckets.size === 0) return;
+    if (buckets.size === 0) return false;
     type PendingItem = {
       key: string;
       seq: number;
@@ -725,7 +725,7 @@ export function InspectorProvider({
       pendingRef.current = new Map();
       setPendingCount(0);
       history.clear();
-      return;
+      return false;
     }
     setCommitting(true);
     try {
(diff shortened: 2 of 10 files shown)


fatima-n09:
@coderabbitai review

fatima-n09:
@coderabbitai full review

stantheman0128:
Yield to #429. Same #408. They throw `NoOpEditError` and stop the Save bar success flash; this PR does not. Keep the locale hunk if 429 lands without zh-TW strings.

fatima-n09:
> Yield to #429. Same #408. They throw `NoOpEditError` and stop the Save bar success flash; this PR does not. Keep the locale hunk if 429 lands without zh-TW strings.

Thanks for the feedback! I was already working on extending this PR to cover the Save bar success flash just pushed the update.
    
commitEdits now returns a boolean so SaveCard only shows the success flash when the file actually changed. This keeps the approach return-value-based rather than throwing, which avoids try/catch flow control. The no-op toast + pending edits staying buffered were already in the previous commits.

stantheman0128:
Thanks. The Save bar flash is covered here now.

#429 still throws NoOpEditError on the batch path, and it has an e2e that fails if changed:false still clears the buffer. This PR still has the locale keys.

I would still take #429 unless this branch also adds those two. Keep the zh-TW / zh-CN / ja hunk either way.

PR #429, opened 2026-08-28
fix(core): warn when batch edit does not change the file

### LINKED ISSUE #408: [Bug]: The batch edit route reports success but does not save the file

PARZIVAL7498:
## Summary
- Treat `/__edit/batch` responses with `changed: false` like the single-edit `NoOpEditError` path: keep pending edits and toast a warning instead of clearing the buffer.
- Stop the Save bar from flashing success when a save rejects (no-op or failed write).
- Add an inspector e2e covering the no-op batch save path.

Fixes #408

## Test plan
- [ ] Open Inspector, change a style (e.g. font size), Save — file updates and Save bar shows success
- [ ] Select the same element, set the same style value again, Save
- [ ] Confirm network response for `/__edit/batch` has `changed: false`
- [ ] Confirm Save bar does **not** show success, pending edits remain, and a warning toast mentions the source file did not change
- [ ] `pnpm check` / typecheck for `@open-slide/core`
- [ ] Optional: `packages/core` inspector e2e `a no-op batch save keeps pending edits and does not show success`

## Summary by CodeRabbit

- **Bug Fixes**
  - Saving inspector edits now detects when the source file is unchanged.
  - Pending edits remain visible instead of being incorrectly marked as saved.
  - Success notifications are suppressed when no changes are applied.
  - Clear notices are shown when the source file has not changed.
  - Failed saves now preserve the unsaved state and report errors accurately.

### DIFF
--- a/.changeset/batch-edit-noop.md
+++ b/.changeset/batch-edit-noop.md
@@ -1,0 +1,5 @@
+---
+'@open-slide/core': patch
+---
+
+Warn and keep pending edits when a batch save does not change the slide file.
--- a/packages/core/e2e/tests/inspector.spec.ts
+++ b/packages/core/e2e/tests/inspector.spec.ts
@@ -57,6 +57,39 @@ test.describe('inspector editing', () => {
     await expect.poll(() => readSlideSource('insp-save')).toContain('Edited via inspector');
   });
 
+  test('a no-op batch save keeps pending edits and does not show success', async ({
+    page,
+    request,
+  }) => {
+    await openEditable(page, request, 'insp-noop');
+    await page.getByTitle('Inspect').click();
+    await editorCanvas(page).getByText('Editable headline').click();
+
+    const textField = page.locator('aside[data-inspector-ui]').getByPlaceholder('Element text');
+    await textField.fill('No-op target');
+    const firstSave = page.waitForResponse(
+      (res) => res.url().includes('/__edit/batch') && res.request().method() === 'POST',
+    );
+    await page.getByRole('button', { name: 'Save' }).click();
+    expect((await firstSave).status()).toBe(200);
+    await expect.poll(() => readSlideSource('insp-noop')).toContain('No-op target');
+
+    await editorCanvas(page).getByText('No-op target').click();
+    await textField.fill('No-op target');
+    await expect(page.getByText('1 unsaved change')).toBeVisible();
+
+    const noopSave = page.waitForResponse(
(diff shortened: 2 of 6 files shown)


stantheman0128:
Support. Same #408 as #421. This is the fuller vehicle: throw the existing `NoOpEditError`, `Promise.allSettled` so the Save bar does not flash success, and an e2e that goes red if `changed: false` still clears the buffer.

#421 has the zh-TW / zh-CN / ja keys. `NoOpEditError` is hardcoded English, so a zh-TW inspector still toasts English. Fold those strings onto this PR rather than merging both.

`changed !== false` is the right default if an older server omits the field. #421's `?? false` would toast a no-op on every save there.

PR #430, opened 2026-08-28
fix(core): select tagged text inside Agenda-style wrappers

### LINKED ISSUE #412: [Bug]: Inspector cannot select text inside the Agenda primitive
### Summary

Inspect mode activates, but clicking text rendered by the `Agenda` primitive does not produce a selection frame or open the inspector panel. The same inspector works for directly source-tagged text elements on other pages.

### Steps to reproduce

1. Use `@open-slide/core@1.18.0` in a slide project.
2. Render a page containing `Agenda` with inline children, for example:

   ```tsx
   <Agenda>
     <span>What changed?</span>
     <span>What did it expose?</span>
   </Agenda>
   ```

3. Run `pnpm dev` and open that page in the editor.
4. Click `Inspect`.
5. Click one of the agenda row's text spans.

### Expected behaviour

Clicking the agenda text should select the source-backed element, show the blue selection frame, and open the Inspect panel with its source location and editable content.

### Actual behaviour

Inspect toggles on, but clicking the agenda text leaves the inspector panel closed and no selection frame appears. DOM inspection shows `data-slide-loc` on the inner `<span>`, but the inspector promotes the inline span to the containing `<li>` before calling source resolution. `findSlideSource(..., { hostOnly: true })` then cannot resolve the slide source from that `<li>`.

### Affected package

@open-slide/core

### Version

@open-slide/core@1.18.0

### Environment

Node v24.18.0, pnpm 10.4.1, Windows 11, Microsoft Edge 151.0.0.0

### Additional context

Reproduced in an Aurora demo deck. Page 1 inspector works; page 2 fails because `Agenda` renders `<li>` containers around source-tagged spans. No runtime exception was observed.

This is related in theme to #65, but appears distinct: source tags are present on the rendered elements; the failure happens when the inspector promotes the tagged inline span to its containing `<li>` before source resolution.

The relevant interaction appears to be between `pickInspectorTarget()` in `src/app/components/inspector/inspect-overlay.tsx`, which climbs inline text to an editable container, and `findSlideSource()` in `src/app/lib/inspector/fiber.ts`, which then starts source resolution from the promoted element.

A likely fix is to preserve a tagged source element when promoting editable text containers, or resolve `data-slide-loc` before/after promotion. A workaround is to use direct host elements as the inspector anchor.

PARZIVAL7498:
## Summary
- `pickInspectorTarget()` was promoting source-tagged inline text (e.g. `<span data-slide-loc>`) up into untagged Agenda/list wrappers (`<li>`).
- `findSlideSource()` only walks ancestors via `closest('[data-slide-loc]')`, so selection silently failed with no frame or panel.
- Keep the tagged inline when the editable container has no `data-slide-loc`; still promote marks into a tagged `<p>` as before.
- Add unit coverage for the Agenda-style case.

Fixes #412

## Test plan
- [ ] Render a slide with an Agenda-style helper that wraps slide `span` children in `<li>`
- [ ] Enable Inspect and click agenda row text
- [ ] Confirm blue selection frame + Inspect panel open with source loc and editable text
- [ ] Confirm normal `<p>a <strong>b</strong></p>` still selects the paragraph
- [ ] `pnpm test` (`pick-target.test.ts`) and `pnpm check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved inspector selection for tagged inline text inside untagged Agenda-style containers.
  * Preserved the most specific tagged inline element when selecting nested text.
  * Safely handles cases where no starting element is available.

* **Tests**
  * Added coverage for inline tag selection, nested marks, and empty selections.

### DIFF
--- a/.changeset/inspector-agenda-pick.md
+++ b/.changeset/inspector-agenda-pick.md
@@ -1,0 +1,5 @@
+---
+'@open-slide/core': patch
+---
+
+Keep inspector selection on tagged inline text inside untagged Agenda-style wrappers.
--- a/packages/core/src/app/lib/inspector/pick-target.test.ts
+++ b/packages/core/src/app/lib/inspector/pick-target.test.ts
@@ -1,0 +1,102 @@
+import { afterAll, beforeAll, describe, expect, it, vi } from 'vitest';
+import { pickInspectorTarget } from './pick-target.ts';
+
+class FakeText {
+  readonly nodeType = 3;
+  constructor(readonly textContent: string) {}
+}
+
+class FakeHTMLElement {
+  parentElement: FakeHTMLElement | null = null;
+  childNodes: Array<FakeHTMLElement | FakeText> = [];
+  dataset: Record<string, string> = {};
+  private root: FakeHTMLElement | null = null;
+
+  constructor(readonly tagName: string) {}
+
+  get textContent(): string {
+    return this.childNodes
+      .map((n) => (n instanceof FakeText ? n.textContent : n.textContent))
+      .join('');
+  }
+
+  hasAttribute(name: string): boolean {
+    return name === 'data-slide-loc' && this.dataset.slideLoc !== undefined;
+  }
+
+  closest(selector: string): FakeHTMLElement | null {
+    if (selector !== '[data-inspector-root]') return null;
+    return this.root;
+  }
+
+  contains(other: FakeHTMLElement): boolean {
+    for (let cur: FakeHTMLElement | null = other; cur; cur = cur.parentElement) {
+      if (cur === this) return true;
+    }
+    return false;
+  }
+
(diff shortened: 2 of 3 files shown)


stantheman0128:
Same #412 hole, after #424 extracted `pickInspectorTarget`. Support that layer.

The keep still promotes a tagged inline into a tagged wrapper host. Walking past the loc-tagged span, then hitting `<div data-slide-loc>` with only inline children, returns the host because taggedInline is only preferred when the container has no loc. Agenda wrappers are untagged `<li>`, so that path works. The tagged-wrapper steal is the case #428 returns on immediately via `isTaggedInlineText`.

This PR is unit tests only. #428 also has the Agenda e2e for that keep. I can rebase #428 on request.

PR #431, opened 2026-08-28
fix(core): slide toolbar back button follows browser history and preserves last home location

benben6515:
## Problem

The toolbar back button (←) on the slide route is a hard `<Link to="/">`. It always lands on the bare home route, which:

- discards query state on the home browser — e.g. a folder filter like `/?f=draft` silently resets to `/`
- ignores where the user actually came from (themes gallery, assets page)

Going "back" is expected to behave like the browser's back, especially since the home URL carries view state in its query.

## Change

The back button now follows browser-history semantics:

1. If there is an in-app history entry to return to (react-router records its entry index in `history.state`; `idx > 0`), it calls `navigate(-1)` — the exact previous location, query intact.
2. Otherwise (slide URL opened directly, fresh tab / deep link), it falls back to the **last remembered home location**, kept in `sessionStorage` by `HomeShell` (which owns `/`, `/themes`, `/assets`), defaulting to `/`.

The element changes from `<Link>` to a `<button>` (role change is covered by an updated e2e test). Page-level navigation inside a deck is untouched — it still uses `replace: true`.

| Scenario | Before | After |
| --- | --- | --- |
| `/?f=draft` → open slide → ← | `/` | `/?f=draft` |
| Themes gallery → open slide → ← | `/` | Themes gallery |
| Deep link to `/s/foo` → ← | `/` | last visited home location (default `/`) |

## Testing

- `pnpm check`, `pnpm typecheck`, `pnpm test` (313 unit tests) — all pass
- `playwright test viewer` — 14 pass, including two new cases:
  - back returns to the previous browser location with its query intact
  - deep-link fallback lands on the last home location, query included
  - (plus the existing back-link test updated from the `link` to the `button` role)
- Manually exercised in `apps/demo`: home → slide → back, folder-filtered home → slide → back, and cold deep-link fallback

## Changeset

Included: `@open-slide/core` patch.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the slide toolbar’s “Back to home” behavior.
  - Preserves home filters and query parameters when navigating back.
  - Returns to the previous browser location when available.
  - Falls back to the most recent home location when no navigation history exists.
- **Tests**
  - Added coverage for standard, direct-entry, and fallback navigation scenarios.

### DIFF
--- a/.changeset/slide-back-browser-history.md
+++ b/.changeset/slide-back-browser-history.md
@@ -2,4 +2,4 @@
 "@open-slide/core": patch
 ---
 
-Return the slide toolbar back button to the previous browser location, preserving home query state like folder filters.
+Slide toolbar back button returns to the previous browser location and preserves home query state like folder filters.
--- a/.changeset/slide-back-browser-history.md
+++ b/.changeset/slide-back-browser-history.md
@@ -1,0 +1,5 @@
+---
+"@open-slide/core": patch
+---
+
+Return the slide toolbar back button to the previous browser location, preserving home query state like folder filters.
--- a/packages/core/e2e/tests/viewer.spec.ts
+++ b/packages/core/e2e/tests/viewer.spec.ts
@@ -90,11 +90,14 @@ test.describe('slide viewer', () => {
   });
 
   test('back returns to the previous browser location with its query intact', async ({ page }) => {
-    await page.goto('/');
+    await page.goto('/?f=draft');
+    await page.waitForFunction(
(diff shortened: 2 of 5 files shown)


benben6515:
Thanks for the review @coderabbitai — both nitpicks applied in 9fe2df9: the changeset now uses present tense, and the history-path e2e test now exercises query preservation from a filtered home view (`/?f=draft`) instead of the bare home route.

1weiho (approved):

1weiho:
@benben6515 Hi 👋🏻
謝謝🤍

PR #433, opened 2026-08-31
fix(core): only tag the deck entry for inspector edits

SegawaBeer:
## What

`data-slide-loc` tags carry a bare `line:column`, and `/__edit` resolves every write to `slides/<id>/index.tsx`. The plugin tagged every nested `.tsx` under a deck folder, so a tag minted in an imported file described a position in a file the write path never opens. `findInnermostJsxElement` then fell back to whichever element in the entry file happened to span that position and edited it.

Reproduce with a deck whose `index.tsx` imports pages from a sibling file (e.g. `./_shared/pages.tsx`): clicking any of the imported JSX in the inspector saves an edit into an unrelated element of `index.tsx`. The longer the two files, the more likely the fallback finds a live target instead of failing.

## Fix

Tag only the deck entry, `<slidesDir>/<id>/index.tsx`.

Nothing editable is lost. `findSlideSource`'s fiber fallback already accepts a `_debugSource` only when it ends in `/slides/<id>/index.tsx`, on the stated assumption that imported component files are untransformed — they were not, and the tag path runs first, so that guard never got a say. Untagged JSX now reaches the fallback and is reported as not targetable, which is the correct answer for an element `/__edit` cannot write.

## Tested

- `pnpm test` — 306 passed, 21 files. `loc-tags-plugin.test.ts` (25 tests) had its three "tags nested/shared/numbered file" cases inverted to assert the transform is skipped, plus a new case for a path that merely ends in `index.tsx`.
- `pnpm typecheck` and `pnpm check` clean.
- Changeset included (`patch` for `@open-slide/core`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Inspector edits now apply location tags only to slide entry files.
  * Imported JSX, sibling files, numbered slides, and nested non-entry files are no longer incorrectly tagged.
  * Improved path handling across nested folders and Windows-style paths.

### DIFF
--- a/.changeset/entry-only-loc-tags.md
+++ b/.changeset/entry-only-loc-tags.md
@@ -1,0 +1,5 @@
+---
+"@open-slide/core": patch
+---
+
+Only tag the deck's `index.tsx` for inspector edits, so clicking JSX imported from a sibling file no longer writes to the wrong place in the entry file.
--- a/packages/core/src/vite/loc-tags-plugin.test.ts
+++ b/packages/core/src/vite/loc-tags-plugin.test.ts
@@ -137,16 +137,30 @@ describe('locTagsPlugin', () => {
     expectTaggedTransform('/repo/slides/cover/index.tsx');
   });
 
-  it('tags shared slide source files', () => {
-    expectTaggedTransform('/repo/slides/cover/shared.tsx');
+  /**
+   * These three used to be tagged, and that was the bug. A tag is a bare
+   * `line:column`; `/__edit` always writes to `slides/<id>/index.tsx`. So a tag
+   * minted in a sibling file sent the inspector a position in a file it would
+   * never open, and `findInnermostJsxElement`'s containment fallback then edited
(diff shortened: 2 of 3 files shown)


stantheman0128:
Support. A loc tag is a bare `line:column` and `/__edit` always writes `slides/<id>/index.tsx`. Tagging a sibling was a lying loc; the containment fallback then edited whatever in the entry spanned that position. Skipping non-entry files is the write-path invariant `findSlideSource` already encoded for `_debugSource`.

This does not replace #367. After sibling files are untagged, selecting imported JSX still needs the call-site fiber walk (drop `hostOnly` at inspector). Complementary. Keeping #367 open.

stantheman0128:
Follow-up on the lying sibling loc: I opened https://github.com/1weiho/open-slide/pull/441 so `data-slide-loc` names the source file and `/__edit` writes there. If that lands, the entry-only untag here is no longer needed. Still support this as a stopgap until then.

PR #439, opened 2026-09-02
fix(web): make landing Live Demo keyboard-focusable

### LINKED ISSUE #425: [UI]: Make Live Demo Keyboard-Navigable on Landing page.
### Summary

As a keyboard user, I want to navigate the Live Demo using my keyboard so that I can interact with the embedded presentation without relying only on mouse interaction.

The Live Demo already contains keyboard navigation logic for actions such as ArrowLeft, ArrowRight, Home, and End. However, the slide player container is not keyboard-focusable, which prevents users from reliably accessing the existing keyboard navigation functionality.

### Steps to reproduce

- Open the Open-Slide landing page.
- Scroll to the Live Demo section.
- Click somewhere outside the Live Demo.
- Press Tab repeatedly until reaching the Live Demo.
- Try using ArrowRight or ArrowLeft to navigate between slides.
- Compare the behavior with clicking the visible previous/next controls.

### Expected behaviour

As a keyboard user, I should be able to:

Reach/focus the Live Demo on Landing page using Tab.
Use the existing keyboard controls to navigate between slides.
Use ArrowRight to move to the next slide.
Use ArrowLeft to move to the previous slide.
Use Home to move to the first slide.
Use End to move to the last slide.

The existing mouse-based navigation should continue to work exactly as it does today.

### Actual behaviour

The slide player container is not keyboard-focusable.

As a result, keyboard users cannot reliably access the existing keyboard navigation functionality through normal Tab navigation.

Mouse navigation continues to work.

### Affected package

scaffolded project / landing-page Live Demo section.

### Version

Current development version..

### Environment

Windows ,  Chrome

### Additional context

The issue is in components/landing/inline-slide-player.tsx. The component already implements keyboard navigation through onKeyDown for keys such as ArrowLeft, ArrowRight, Home, and End, but the root <div> is not keyboard-focusable.

```
Before 
<div
  ref={rootRef}
  onKeyDown={onKeyDown}
  aria-roledescription="slide player"
  aria-label={`Slide ${index + 1} of ${count}`}
>
After
<div
  ref={rootRef}
  tabIndex={0}
  onKeyDown={onKeyDown}
  aria-roledescription="slide player"
  aria-label={`Slide ${index + 1} of ${count}`}
>

The initial fix should be limited to adding:
tabIndex={0}
```

This should make the existing keyboard navigation accessible without changing the current slide navigation, mouse controls, animations, or other behavior.

As shown in the screenshot, the slider works with mouse interaction rather than keyboard input. if i press left and right arrow button, will not move. 

Image

ssk090:
Add tabIndex={0} so Tab can focus the Live Demo player. Existing arrow/Home/End handlers then work. Fixes #425.

## Summary by CodeRabbit

* **Accessibility**
  * Made the inline slide player keyboard-focusable, enabling keyboard navigation when it receives focus.

### DIFF
--- a/apps/web/components/landing/inline-slide-player.tsx
+++ b/apps/web/components/landing/inline-slide-player.tsx
@@ -59,6 +59,7 @@ export function InlineSlidePlayer({ index, onIndexChange }: Props) {
   return (
     <div
       ref={rootRef}
+      tabIndex={0}
       onKeyDown={onKeyDown}
       aria-roledescription="slide player"
       aria-label={`Slide ${index + 1} of ${count}`}


1weiho (approved):

1weiho:
@ssk090 Hi 👋🏻

Issue #452, opened 2026-09-07
[Bug]: Inline text edit fails to save for prop-driven text reused across multiple component call sites


kywk:
### Summary

When a presentational component that renders an editable text via a passed-in prop (e.g. `<div>{title}</div>`) is used **multiple times** on the same page, inline-editing that text in the dev-UI inspector can fail to save with:

```
Couldn't save: line <N>: no text candidate matches the current value
```

The framework disambiguates which call site is being edited by matching the editor's `prevText` against every `title=` string literal collected from the component's call sites (`collectPropCallSiteCandidates` → `buildTextSplice` in `@open-slide/core`, `config-*.js`). If `prevText.trim()` doesn't exactly equal any candidate literal, the save is rejected. In practice this is easy to trigger: retrying an edit on top of a partially-changed value, or any rendered-vs-source whitespace difference (e.g. full-width vs half-width space), leaves no candidate matching and the edit becomes unsaveable.

### Steps to reproduce

1. Scaffold a slide with a reusable text component used more than once, e.g.:
   ```tsx
   const Card = ({ title }: { title: string }) => (
     <div style={{ fontSize: 25 }}>{title}</div>
   );
   const Page = () => (
     <div>
       <Card title="Alpha" />
       <Card title="Beta" />
     </div>
   );
   ```
2. `pnpm dev`, open the deck, toggle the inspector.
3. Click the "Alpha" text and start editing it inline.
4. Change a few characters (or retry after a prior partial edit) and save.

### Expected behaviour

The inline edit is saved to the correct call-site `title=` literal. Ideally the inspector identifies the target call site by position/index (which instance was clicked) rather than only by matching the pre-edit text value, so an in-progress edit or a benign whitespace difference doesn't make the text unsaveable.

### Actual behaviour

Save is rejected with `Couldn't save: line <N>: no text candidate matches the current value`, where line N points at the `{title}` JSX expression inside the component definition (not the call site). The text can no longer be saved via the inspector; the only reliable workaround is to edit the `title="…"` source literal by hand, or reload the page and edit in a single clean pass so `prevText` still matches the source literal exactly.

### Affected package

`@open-slide/core`

### Version

`@open-slide/core@2.0.0-beta.1`

### Environment

Node 22.23.2, pnpm 10.17.0, macOS 26.6.2

### Additional context

[...]

PR #450, opened 2026-09-06
feat(core): add visual slide editing and arrangement tools

1weiho:
Slides can now be edited directly on the canvas with dragging, eight resize handles, rotation, multi-selection, marquee selection, smart snapping guides, alignment, distribution, layer controls, and precise position and size inputs.

Development opens in edit mode: click to select or move an object, double-click or press Enter to edit text, and use Escape to return to object selection. A persistent Format sidebar shows contextual Text, Image, Style, and Arrange controls without shifting the canvas as selection changes. Preview restores slide interactions and keyboard navigation; Format and Design share the same sidebar space.

Canvas text selection and sidebar formatting use the same range and undo/redo history. Text edits and range formatting save in their original order, preserving formatting after further typing, undo/redo, and reload. Each arrangement gesture is one undo/redo step. Includes localized controls, documentation, and changesets.

Gesture cursors remain visible over slide content. Text replay retains inherited and nested whitespace context. Failed text steps stop their dependent steps on the server while independent edits continue; retry and discard preserve successful changes and recover pending instances after source reloads. Undo explicitly restores an edit's original sequence position.

Repeated instances from one JSX definition require selecting a unique parent to arrange independently. Inline text uses its parent container, and layer controls follow CSS stacking contexts.

Mixed CSS whitespace modes and significant spaces beside `<br>` still have existing source-save normalization limits; their preview, undo/redo, and discard behavior are covered here. Saving and reloading are also covered for inherited `pre`, `pre-wrap`, and `break-spaces`.

Validation:
- 360 unit tests passed.
- 50 related browser cases passed across editor UX, inspector, visual editor, history, whitespace replay, and partial-save recovery, including 10 new review regressions.
- Core TypeScript and package build passed.
- Full repository Biome and staged diff checks passed.
- Manually verified selection, text formatting, stable sidebar layout, and Preview keyboard navigation in the demo.

## Summary by CodeRabbit

* **New Features**
  * Added visual slide editing with dragging, resizing, rotation, multi-selection, snapping, alignment, distribution, and layer controls.
  * Added Format and Arrange panels, Preview/Edit modes, inline text editing, image replacement, keyboard nudging, and undo/redo and save shortcuts.
  * Added batch editing with per-edit results and error handling.
* **Bug Fixes**
  * Improved edit recovery, whitespace preservation, gesture feedback, and history reliability.
* **Documentation**
  * Expanded visual editor guidance and shortcuts.
* **Localization**
[...]

PR #455, opened 2026-09-09
fix(core): preserve original text when editing reused components

NickYCLin:
﻿Inline typing captured `prevText` after the browser had already changed the DOM. With multiple `<Card title="..." />` call sites, saving then searched for the edited value in the original source and failed. Capture text before input, retain it for save/discard/undo, and keep the initial value through IME composition and failed-save retries.

Match whitespace the same way as the inspector's text reader. Multiple candidates with the same normalized text still fail safely rather than selecting a sibling arbitrarily. This does not introduce positional call-site identity.

Refs #452. The text-snapshot changes overlap with part of #450; this is a focused fix against current main, with whitespace matching and regression coverage.

Validation:
- Full Biome 2.5.7 check passed using the existing cached executable; checkout line endings were normalized for the check, then unrelated files were restored.
- `git diff --check` and Node syntax checks passed.
- Added four unit cases and three browser cases for whitespace, retry, discard, and composition.
- Unit tests, typecheck, and browser tests have not run: GitHub CI run 34316148367 requires maintainer approval (`action_required`). Local dependencies are absent and no packages or browsers were installed. Vercel preview checks also require team authorization.


pablospe:
Hi @NickYCLin, thanks for this fix. I opened #487, which builds on your PR: it keeps your `previousTextRef` capture, the `compositionstart` reset, the provider `op.prevText` fallback and your tests, and credits you as co-author. On top of that it applies the same whitespace collapsing in `elementTextCandidateMatches` (used by range-style edits), adds a regression test for that path, and adjusts two e2e cases that were passing on main only because their first keystroke was a space. Could you take a look at #487? If it covers what you intended here, would you be OK closing this one in favour of it? Happy to adjust anything.

